PR #8, opened 2020-04-13
digest image name support

FrimIdan:
Report scan failure when job creation failed
Fix bug in job name creation when image name used digest and not tag

### DIFF
--- a/pkg/orchestrator/job-manager.go
+++ b/pkg/orchestrator/job-manager.go
@@ -109,7 +109,7 @@ func getSimpleImageName(imageName string) string {
 	repoEnd := strings.LastIndex(imageName, "/")
 	imageName = imageName[repoEnd+1 :]
 
-	digestStart := strings.LastIndex(imageName, "@sha256:")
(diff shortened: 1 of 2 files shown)


FrimIdan:
@akpsgit please review

pkg/orchestrator/job-manager.go
```diff
@@ -104,11 +107,20 @@ func (o *Orchestrator) runJob(data *scanData) error {
 // Due to K8s names constraint we will take the image name w/o the tag and repo
 func getSimpleImageName(imageName string) string {
 	repoEnd := strings.LastIndex(imageName, "/")
-	imageName = imageName[repoEnd+1 : len(imageName)]
+	imageName = imageName[repoEnd+1 :]
+
+	digestStart := strings.LastIndex(imageName, "@sha256:")
```
akpsgit, line 112: From looking at the format [here](https://kubernetes.io/docs/concepts/configuration/overview/#container-images), looks like it's `<image-name>@<digest>` 
Don't know if we should assume sha256

PR #9, opened 2020-04-21
get image name from pod.Spec.Containers and image id from pod.Status.…

FrimIdan:
…ContainerStatuses


FrimIdan:
@akpsgit

akpsgit:
@FrimIdan, looks good, please merge.

PR #11, opened 2020-04-27
Use proxy configuration in the scanner job

FrimIdan:


FrimIdan:
@akpsgit

pkg/config/config.go
```diff
@@ -14,35 +15,56 @@ const (
...
-	viper.SetDefault(KlarTrace, "false")                   // Run Klar in more verbose mode
+	viper.SetDefault(KlarTrace, "false") // Run Klar in more verbose mode
 	viper.SetDefault(ClairAddress, "clair.kubei")
 
 	viper.AutomaticEnv()
 }
 
+// extracts the service address from the full path.
+// example: `http://kubei.kubei:8081/result/` should return `kubei.kubei`
+func getServiceAddress(serviceFullPath string) string {
```
akpsgit, line 40: This can be done using url.Parse("http://kubei.kubei:8081/result/"):
	u, err := url.Parse("http://kubei.kubei:8081/result/")
	if err != nil {
		....
	}
	fmt.Println(u.Hostname())

pkg/config/config_test.go
```diff
@@ -0,0 +1,64 @@
+package config
+
+import "testing"
+
+func Test_getServiceAddress(t *testing.T) {
```
akpsgit, line 5: No need

pkg/config/config.go
FrimIdan: Thanks, from url.Parse() comments	
```
Trying to parse a hostname and path without a scheme is invalid
 but may not necessarily return an error, due to parsing ambiguities.
```

I will add a validation that the path includes scheme and use url.Parse()

FrimIdan:
@akpsgit I've made proxy configuration to effects only on scanner jobs and not Kubei.
Added README notes and fix the url.Parse() issue.

README.md
```diff
@@ -66,6 +66,9 @@ The file `deploy/kubei.yaml` is used to deploy and configure Kubei on your clust
 
 ![](images/kubei-results.png)     
 
+## Running Kubei behind Proxy
```
akpsgit, line 69: @FrimIdan Running Kubei with an external HTTP/HTTPS proxy

README.md
```diff
@@ -66,6 +66,9 @@ The file `deploy/kubei.yaml` is used to deploy and configure Kubei on your clust
...
+
+Uncomment proxy env variables configuration under Clair and Kubei deployments in `deploy/kubei.yaml`.
```
akpsgit, line 71: Uncomment and configure the proxy env variables for the Clair and Kubei deployments in `deploy/kubei.yaml`.

PR #17, opened 2020-06-17
handle images with tag and digest

FrimIdan:


pkg/scanner/job-manager.go
```diff
@@ -179,22 +180,17 @@ func (s *Scanner) deleteJob(job *batchv1.Job) {
...
-	// remove tag if exists
-	if tagStart != -1 {
-		return imageName[:tagStart]
-	}
+	refName := ref.Name()
+	// Take only image name from repo path (ex. solsson/kafka ==> kafka)
+	repoEnd := strings.LastIndex(refName, "/")
 
-	return imageName
+	return refName[repoEnd+1:]
```
akpsgit, line 193: ParseNormalizedNamed should guard us, but maybe worth checking that we don't crash on imageName that ends with /

FrimIdan: ParseNormalizedNamed will failed to parse when it's an invalid reference format.
I've change the code to return error when ParseNormalizedNamed() fails, if we can't normalized it the job creation will probably failed so better to fail earlier.

PR #53, opened 2021-05-03
point to klar with stripped commands

akpsgit:


awkwardFish (approved):

deploy/kubei.yaml
```diff
@@ -202,11 +202,11 @@ spec:
             done
       containers:
       - name: kubei
-        image: gcr.io/eticloud/k8sec/kubei:1.0.11
+        image: gcr.io/eticloud/k8sec/kubei:1.0.12
```
FrimIdan, line 205: @akpsgit  why kubei need new tag?

akpsgit: not really needed. this commit (nor the previous) related to user experience in kubei

FrimIdan: so we should not create new tag here, right?

FrimIdan (approved):

PR #126, opened 2022-04-12
Fix getting local image hash from repo digest

pbalogh-sa:


shared/pkg/analyzer/syft/syft.go
```diff
@@ -51,10 +52,12 @@ func New(conf *config.Config,
...
 		config:     config.CreateSyftConfig(conf.Analyzer, conf.Registry),
 		resultChan: resultChan,
+		localImage: conf.LocalImageScan,
 	}
 }
 
-func (a *Analyzer) Run(sourceType utils.SourceType, src string) error {
+func (a *Analyzer) Run(sourceType utils.SourceType, userInput string) error {
+	src := utils.SetSource(a.localImage, sourceType, userInput)
 	src = utils.CreateUserInput(sourceType, src)
```
FrimIdan, line 61: it looks like `utils.CreateUserInput` and `utils.SetSource` should be combined into a single function, am I wrong?
What is the expected input in `CreateResult`? The source? The modified user input? The original user input? We need to better understand what is used where in both syft and grype.

cli/cmd/scan.go
```diff
@@ -170,7 +169,7 @@ func vulnerabilityScanner(cmd *cobra.Command, args []string) {
 		logger.Fatalf("Failed to present results: %v", err)
 	}
 
-	layerCommands, err := getLayerCommandsIfNeeded(sourceType, src, appConfig.SharedConfig.Registry)
+	layerCommands, err := getLayerCommandsIfNeeded(sourceType, args[0], appConfig.SharedConfig.Registry)
```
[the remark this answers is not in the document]
pbalogh-sa: ack, thx

shared/pkg/analyzer/syft/syft.go
pbalogh-sa: You're right, It can be one function. In the case of the input type is an image we have to extend the image soruce with docker or registry for syft and grype. I will update it.

pbalogh-sa: The expected input should be the `userInput` in the case of `CreateResults`.

shared/pkg/utils/image_helper/image_helper.go
```diff
@@ -146,20 +147,15 @@ func stripDockerMetaFromCommand(command string) string {
 	return ret
 }
 
-func getV1Image(imageName string, registryOptions *image.RegistryOptions) (containerregistry_v1.Image, error) {
-	source, location, err := image.DetectSource(imageName)
-	if err != nil {
-		return nil, fmt.Errorf("falied to detect source for image: %s. %v", imageName, err)
-	}
+func getV1Image(imageName string, registryOptions *image.RegistryOptions, localImage bool) (containerregistry_v1.Image, error) {
 	log.Debugf("pulling image info directly from registry image=%q", imageName)
```
FrimIdan, line 152: `log.Debugf("pulling image info directly from registry image=%q", imageName)` - command should be under the `default` switch case, no?

shared/pkg/utils/input.go
```diff
@@ -41,9 +41,16 @@ func ValidateInputType(inputType string) (SourceType, error) {
 	}
 }
 
-func CreateUserInput(sourceType SourceType, src string) string {
+func CreateSource(sourceType SourceType, src string, localImage bool) string {
 	if sourceType != IMAGE {
 		src = fmt.Sprintf("%s:%s", sourceType, src)
```
FrimIdan, line 46: We should return here and not continue to `setImageSource`. (or use a switch case on `sourceType`)
Better to add a unit test here.

pbalogh-sa: oops, thx , I will add unit test

shared/pkg/utils/image_helper/image_helper.go
pbalogh-sa: yes

FrimIdan (approved):

shared/pkg/utils/input_test.go
```diff
@@ -0,0 +1,110 @@
...
+			name: "local image source",
+			args: args{
+				sourceType: DIR,
+				localImage: true,
+				src:        "test/path",
+			},
+			want: "dir:test/path",
+		},
+		{
+			name: "remote image source",
```
akpsgit, line 95: remote file source?

PR #142, opened 2022-04-28
feat: CIS docker benchmarks results backend + api

FrimIdan:


backend/pkg/rest/runtime_scan_controller.go
```diff
@@ -81,9 +83,12 @@ func (s *Server) GetRuntimeScanResults(params operations.GetRuntimeScanResultsPa
 	// in the case of no application IDs to look by, nothing was scanned, so we can just return
 	if len(state.runtimeScanApplicationIDs) == 0 {
 		return operations.NewGetRuntimeScanResultsOK().WithPayload(&models.RuntimeScanResults{
-			Counters:                 &models.RuntimeScanCounters{},
-			Failures:                 failures,
-			VulnerabilityPerSeverity: []*models.VulnerabilityCount{},
+			CisDockerBenchmarkCountPerLevel: []*models.CISDockerBenchmarkLevelCount{},
+			CisDockerBenchmarkCounters:      &models.CISDockerBenchmarkScanCounters{},
+			CisDockerBenchmarkScanEnabled:   quickScanConfig.CisDockerBenchmarkScanEnabled,
```
akpsgit, line 88: it's not correct to show whether CIS benchmark was enabled in the last runtime scan by fetching it from config, because it can be set post-scan.

backend/pkg/rest/runtime_scan_controller.go
```diff
@@ -106,10 +111,32 @@ func (s *Server) GetRuntimeScanResults(params operations.GetRuntimeScanResultsPa
...
+			WithPayload(oopsResponse)
+	}
+
 	return operations.NewGetRuntimeScanResultsOK().WithPayload(&models.RuntimeScanResults{
-		Counters:                 counters,
-		Failures:                 failures,
-		VulnerabilityPerSeverity: vulPerSeverity,
+		CisDockerBenchmarkCountPerLevel: cisDockerBenchmarkCountPerLevel,
+		CisDockerBenchmarkCounters:      cisDockerBenchmarkCounters,
+		CisDockerBenchmarkScanEnabled:   quickScanConfig.CisDockerBenchmarkScanEnabled,
```
akpsgit, line 136: it's not correct to show whether CIS benchmark was enabled in the last runtime scan by fetching it from config, because it can be set post-scan.

backend/pkg/rest/runtime_scan_controller.go
```diff
@@ -158,6 +207,21 @@ func (s *Server) getRuntimeScanCounters(filters *database.CountFilters) (*models
 	}, nil
 }
 
+func (s *Server) getRuntimeScanCisDockerBenchmarkCounters(filters *database.CountFilters) (*models.CISDockerBenchmarkScanCounters, error) {
+	appCount, err := s.dbHandler.ApplicationTable().Count(filters)
+	if err != nil {
+		return nil, fmt.Errorf("failed to count applications: %v", err)
+	}
+	resCount, err := s.dbHandler.ResourceTable().Count(filters)
```
akpsgit, line 215: resourceCount

backend/pkg/types/cis_docker_benchmark.go
```diff
@@ -0,0 +1,79 @@
+// Copyright © 2022 Cisco Systems, Inc. and its affiliates.
```
akpsgit, line 1: worth adding unit tests for this file

backend/pkg/database/database.go
```diff
@@ -122,19 +139,29 @@ SELECT applications.*,
        aps.total_high_count,
        aps.total_critical_count,
        aps.highest_severity,
-       aps.lowest_severity
```
akpsgit, line 125: why this is gone?

backend/pkg/rest/runtime_scan_controller.go
FrimIdan: It is fetched from [local state](https://github.com/cisco-open/kubei/blob/d44ddca5fd2a0d4b515a73ccea121408bff808bd/backend/pkg/rest/runtime_scan_controller.go#L74) and not from DB. The local state is being changed only when a [new startScan command arrived](https://github.com/cisco-open/kubei/blob/d44ddca5fd2a0d4b515a73ccea121408bff808bd/backend/pkg/rest/runtime_scan_controller.go#L283). So even if the config was changed post-scan it will not be changed on local state unless there is a new scan.

backend/pkg/database/database.go
FrimIdan: by mistake, will add it back but we don't really need that value.

backend/pkg/rest/runtime_scan_controller.go
akpsgit: oh, I see. Can you please add a comment that it's the last scan config and not the DB config?

akpsgit (approved):

pbalogh-sa (approved):

akpsgit (approved):

PR #144, opened 2022-04-29
Making chart compatible with OpenShift restricted SCC

mtcolman:


mtcolman:
Please note I had to set `version` in Chart.yaml to not be `latest` as I was getting this error at installation time:
```
Error: validation: chart.metadata.version "latest" is invalid
```

FrimIdan:
> Please note I had to set `version` in Chart.yaml to not be `latest` as I was getting this error at installation time:
> 
> ```
> Error: validation: chart.metadata.version "latest" is invalid
> ```

@mtcolman thanks for the contribution. Regarding the `version` change, we are setting it during build & release workflows so you can revert the change there.
https://github.com/cisco-open/kubei/blob/main/.github/workflows/release.yaml#L72-L75

charts/kubeclarity/values.yaml
```diff
@@ -203,12 +219,46 @@ kubeclarity-postgresql:
...
+    enabled: false
+  service:
+    port: 5432
+  resources:
+    limits:
+      cpu: 2
+      memory: 2Gi
+    requests:
+      cpu: 100m
+      memory: 200Mi
```
FrimIdan, line 243: Is that needed? these are the defaults values? I think that the default size is `size: 8Gi` for example.

charts/kubeclarity/values.yaml
```diff
@@ -203,12 +219,46 @@ kubeclarity-postgresql:
...
     runAsUser: 1001
-    runAsNonRoot: true
+  volumePermissions:
+    enabled: true
+    # if using restricted SCC set runAsUser: "auto" and if running under anyuid SCC - runAsUser needs to match the line above
+    securityContext:
+      runAsUser: 1001
+  shmVolume:
+    chmod:
+      enabled: false
```
FrimIdan, line 261: why false?

mtcolman: From postgres chart values.yaml:
>  Init container Security Context
>  @param volumePermissions.securityContext.runAsUser User ID for the init container
>  Note: the chown of the data folder is done to securityContext.runAsUser
>  and not the below volumePermissions.securityContext.runAsUser
>  When runAsUser is set to special value "auto", init container will try to chwon the
>  data folder to autodetermined user&group, using commands: `id -u`:`id -G | cut -d" " -f2`
>  "auto" is especially useful for OpenShift which has scc with dynamic userids (and 0 is not allowed).
>  You may want to use this volumePermissions.securityContext.runAsUser="auto" in combination with
>  pod securityContext.enabled=false and shmVolume.chmod.enabled=false

charts/kubeclarity/values.yaml
```diff
@@ -41,6 +45,10 @@ kubeclarity:
       memory: "1000Mi"
       cpu: "1000m"
 
+  securityContext:
+    runAsGroup: 1000
+    runAsUser: 1000
```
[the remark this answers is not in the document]
mtcolman: Happy with that. Changes made to align.

charts/kubeclarity/values.yaml
mtcolman: Should be `true` by default for alignment with K8s deployments. Code updated and description/comment added.

mtcolman:
@FrimIdan I've made changes, I think fully aligned to your comments (certainly that was the aim). Are you please able to separately resolve the Chart.yaml version=latest issue that is caused by code in the build & release workflows?

charts/kubeclarity/values.yaml
```diff
@@ -203,12 +207,29 @@ kubeclarity-postgresql:
...
+    # Default is true for K8s. Enabled needs to false for OpenShift restricted SCC and true for anyuid SCC
+    enabled: true
+    # fsGroup specification below is not applied if enabled=false. enabled=false is the required setting for OpenShift "restricted SCC" to work successfully.
+    fsGroup: 1001
   containerSecurityContext:
+    # Default is true for K8s. Enabled needs to false for OpenShift restricted SCC and true for anyuid SCC
     enabled: true
+    # runAsUser specification below is not applied if enabled=false. enabled=false is the required setting for OpenShift "restricted SCC" to work successfully.
     runAsUser: 1001
-    runAsNonRoot: true
```
FrimIdan, line 211: why `runAsNonRoot` was removed?

charts/kubeclarity/values.yaml
```diff
@@ -203,12 +207,29 @@ kubeclarity-postgresql:
...
+  volumePermissions:
+    enabled: true
```
FrimIdan, line 225: default is `enabled: false`

FrimIdan:
Thanks @mtcolman few small comments left

Regarding:
> Are you please able to separately resolve the Chart.yaml version=latest issue that is caused by code in the build & release workflows?

Not sure I understand, what is the problem?

charts/kubeclarity/values.yaml
```diff
@@ -203,12 +207,29 @@ kubeclarity-postgresql:
   ## Setup database name and password
   existingSecret: kubeclarity-postgresql-secret
   postgresqlDatabase: kubeclarity
+  # secretKey: postgresql-password
 
-  ## Enable security context
+  serviceAccount:
+    enabled: true
```
FrimIdan, line 213: default is `false` 
https://github.com/helm/charts/blob/master/stable/postgresql/values.yaml#L94

mtcolman: It is good practice to specify a serviceAccount for a deployment, otherwise the `default` serviceAccount is used and this can then lead to someone unknowingly overassigning privilege at a later date. I'd suggest leave this in unless you really want me to remove it...?

charts/kubeclarity/values.yaml
mtcolman: Updated so this is `enabled: false` by default for K8s installs and added description for OpenShift.

charts/kubeclarity/values.yaml
mtcolman: Hadn't removed on purpose.

charts/kubeclarity/values.yaml
FrimIdan: You are right, that is OK - we can keep it. Thanks!

FrimIdan (approved):

mtcolman:
> Thanks @mtcolman few small comments left
> 
> Regarding:
> 
> > Are you please able to separately resolve the Chart.yaml version=latest issue that is caused by code in the build & release workflows?
> 
> Not sure I understand, what is the problem?

Sorry, not an issue, I misread something.

PR #165, opened 2022-05-20
Extending INFO_MESSAGE

oborys:
Adding text "After removing namespace(s), click the 'Start scan' button" in INFO_MESSAGE. As for now, it's not clarified in the hint. And after removing some or all namespaces from the target field, results will be saved only after running new scanning.


pbalogh-sa (approved):

pbalogh-sa:
Thanks @oborys !

PR #178, opened 2022-05-25
Fix image parsing issue in the case of SBOM input.

pbalogh-sa:
If the input is a SBOM file the name and hash of the source will get from the SBOM.

### DIFF
--- a/shared/pkg/converter/cyclonedx_to_syft.go
+++ b/shared/pkg/converter/cyclonedx_to_syft.go
@@ -44,7 +44,7 @@ type emptyMetadata struct{}
 var ErrFailedToGetCycloneDXSBOM = errors.New("failed to get CycloneDX SBOM from file")
 
 func ConvertCycloneDXToSyftJSONFromFile(inputSBOMFile string, outputSBOMFile string) error {
-	cdxBOM, err := getCycloneDXSBOMFromFile(inputSBOMFile)
+	cdxBOM, err := GetCycloneDXSBOMFromFile(inputSBOMFile)
 	if err != nil {
 		return ErrFailedToGetCycloneDXSBOM
 	}
@@ -80,7 +80,7 @@ func saveSyftSBOMToFile(syftBOM syft_sbom.SBOM, outputSBOMFile string) error {
 	return nil
 }
 
-func getCycloneDXSBOMFromFile(inputSBOMFile string) (*cdx.BOM, error) {
+func GetCycloneDXSBOMFromFile(inputSBOMFile string) (*cdx.BOM, error) {
 	inputSBOM, err := os.ReadFile(inputSBOMFile)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read SBOM file %s: %v", inputSBOMFile, err)
--- a/shared/pkg/scanner/grype/common.go
+++ b/shared/pkg/scanner/grype/common.go
@@ -142,7 +142,7 @@ func getSource(doc grype_models.Document, userInput, hash string) scanner.Source
 		hash = image_helper.GetHashFromRepoDigest(doc.Source.Target.(syft_source.ImageMetadata).RepoDigests, userInput)
 		if hash == "" {
 			// set hash using ManifestDigest if RepoDigest is missing
-			hash = doc.Source.Target.(syft_source.ImageMetadata).ManifestDigest
+			hash = strings.TrimPrefix(doc.Source.Target.(syft_source.ImageMetadata).ManifestDigest, "sha256:")
 		}
 	case string:
 		srcName = doc.Source.Target.(string) // nolint:forcetypeassert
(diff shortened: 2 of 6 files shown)


shared/pkg/converter/cyclonedx_to_syft.go
```diff
@@ -121,6 +121,9 @@ func convertCycloneDXtoSyft(bom *cdx.BOM) (syft_sbom.SBOM, error) {
 	if bom.Metadata.Component == nil {
 		return syft_sbom.SBOM{}, fmt.Errorf("cycloneDX metadata component is nil")
 	}
+	if bom.Components == nil {
+		return syft_sbom.SBOM{}, fmt.Errorf("cycloneDX doesent have any components")
```
akpsgit, line 125: cycloneDX doesn't have any components

shared/pkg/scanner/grype/local_grype.go
```diff
@@ -68,13 +68,20 @@ func (s *LocalScanner) run(sourceType utils.SourceType, userInput string) {
...
+	var hash string
+	origInput := userInput
 	if sourceType == utils.SBOM {
 		syftJSONFilePath, cleanup, err := ConvertCycloneDXFileToSyftJSONFile(userInput, s.logger)
 		if err != nil {
 			ReportError(s.resultChan, fmt.Errorf("failed to convert sbom file: %w", err), s.logger)
 			return
 		}
 		defer cleanup()
+		origInput, hash, err = getOriginalInputAndHashFromSBOM(userInput)
```
akpsgit, line 80: do we want to validate that the hash is not empty even if there was no error?

shared/pkg/scanner/grype/remote_grype.go
```diff
@@ -91,8 +91,14 @@ func (s *RemoteScanner) run(sbomInputFilePath string) {
 		return
 	}
 
+	userInput, hash, err := getOriginalInputAndHashFromSBOM(sbomInputFilePath)
+	if err != nil {
+		ReportError(s.resultChan, fmt.Errorf("failed to get original source ahd hash from SBOM: %w", err), s.logger)
```
akpsgit, line 96: ahd -> and

shared/pkg/converter/cyclonedx_to_syft.go
pbalogh-sa: thx

shared/pkg/scanner/grype/remote_grype.go
pbalogh-sa: thx

shared/pkg/scanner/grype/local_grype.go
pbalogh-sa: In the `getSource()` we check the hash is empty or not. If the hash is empty string wi will try to get it from the repodigest or manifestdigest.
https://github.com/openclarity/kubeclarity/blob/f145970f20686b1cde40f5d4eb62991b68b83dae/shared/pkg/scanner/grype/common.go#L133-L146

shared/pkg/scanner/grype/remote_grype.go
```diff
@@ -91,8 +91,14 @@ func (s *RemoteScanner) run(sbomInputFilePath string) {
 		return
 	}
 
+	userInput, hash, err := getOriginalInputAndHashFromSBOM(sbomInputFilePath)
```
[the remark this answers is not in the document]
pbalogh-sa: Please see https://github.com/openclarity/kubeclarity/pull/178#discussion_r887639068

shared/pkg/scanner/grype/common.go
```diff
@@ -117,24 +118,31 @@ func CreateResults(doc grype_models.Document, userInput, scannerName string) *sc
...
+			srcName = userInput
+		}
+		if hash != "" {
+			break
+		}
 		hash = image_helper.GetHashFromRepoDigest(doc.Source.Target.(syft_source.ImageMetadata).RepoDigests, userInput)
 		if hash == "" {
 			// set hash using ManifestDigest if RepoDigest is missing
-			hash = doc.Source.Target.(syft_source.ImageMetadata).ManifestDigest
+			hash = strings.TrimPrefix(doc.Source.Target.(syft_source.ImageMetadata).ManifestDigest, "sha256:")
```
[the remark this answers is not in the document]
pbalogh-sa: ack

shared/pkg/scanner/grype/local_grype.go
akpsgit: so we assume that ```manifestHash := doc.Source.Target.(syft_source.ImageMetadata).ManifestDigest``` will always get something?

pbalogh-sa: If everything is empty we don't have any hash and we can't calculate it as well because we are using sbom as input. Se we let the hash empty.

akpsgit (approved):

shared/pkg/scanner/grype/local_grype.go
pbalogh-sa: The hash can be empty in that case when we are using third party sbom that doesn't care about the hash.

fishkerez (approved):

Issue #226, opened 2022-08-22
DB Queries are really slow


log2:
#### What happened:

Navigation is slow, due to most queries having multi-second timing.

Views should produce quick results, provided that adequate indexes are available, or they are [materialized views](https://www.postgresql.org/docs/current/rules-materializedviews.html) (which could be feasible, since kubeclarity knows exactly when content used to populate views is about to change, i.e., when a scan operation is completed).
 
#### What you expected to happen:

Navigating from page to page should be quick.

#### How to reproduce it (as minimally and precisely as possible):

Populate mentioned tables with hundreds or, better, thousands of tuples. This happens on a cluster I own, which has twenty or more namespaces.

#### Are there any error messages in KubeClarity logs?
(e.g. `kubectl logs -n kubeclarity --selector=app=kubeclarity`)

> kubeclarity
kubeclarity 2022/08/22 15:33:26 /build/backend/pkg/database/application.go:297 SLOW SQL >= 200ms
kubeclarity [6341.415ms] [rows:1] SELECT count(*) FROM "applications_view"
kubeclarity
kubeclarity 2022/08/22 15:33:33 /build/backend/pkg/database/application.go:309 SLOW SQL >= 200ms
kubeclarity [6802.818ms] [rows:50] SELECT * FROM "applications_view" ORDER BY name desc LIMIT 50 OFFSET 250

#### Anything else we need to know?:

I'm using version 2.5.0 of kubeclarity chart.

#### Environment:
- Kubernetes version (use `kubectl version --short`):

>Client Version: v1.24.3
Kustomize Version: v4.5.4
Server Version: v1.21.7

- KubeClarity version (use `kubectl -n kubeclarity exec deploy/kubeclarity -- ./backend version`)
>Version: v2.5.0
Commit: 383f694a7f01c2a47522f520ff3bd4d68b243271
Build Time: 2022-07-18T07:06:08Z
- Cloud provider or hardware configuration:
> Azure (AKS)
- Others:

FrimIdan:
@log2 Thanks for reporting the issue, we will take a look at your suggestions.

PR #240, opened 2022-10-25
Backend: in case of PosrtgreSQL use materialized views 

pbalogh-sa:
Signed-off-by: Peter Balogh <p.balogh.sa@gmail.com>

fixes: #226

### DIFF
--- a/backend/pkg/backend/backend.go
+++ b/backend/pkg/backend/backend.go
@@ -87,8 +87,9 @@ func Run() {
 		go dbHandler.CreateFakeData()
 	}
 
+	dbHandler.SetMaterializedViewHandler()
 	if config.DatabaseDriver == _database.DBDriverTypePostgres {
-		dbHandler.SetMaterializedViewHandler()
+		dbHandler.ViewRefreshHandler.RegisterViewRefreshHandler(_database.RefreshMaterializedViews)
 		go dbHandler.RefreshMaterializedViews()
 	}
 
--- a/backend/pkg/backend/backend.go
+++ b/backend/pkg/backend/backend.go
@@ -87,6 +87,11 @@ func Run() {
 		go dbHandler.CreateFakeData()
 	}
 
+	if config.DatabaseDriver == _database.DBDriverTypePostgres {
+		dbHandler.SetMaterializedViewHandler()
+		go dbHandler.RefreshMaterializedViews()
+	}
+
 	backend := CreateBackend(dbHandler)
 
 	var k8sClientset *kubernetes.Clientset
--- a/backend/pkg/backend/backend.go
+++ b/backend/pkg/backend/backend.go
@@ -85,10 +85,8 @@ func Run() {
 	dbHandler := _database.Init(dbConfig)
 
 	dbHandler.SetMaterializedViewHandler(dbConfig)
-	if config.DatabaseDriver == _database.DBDriverTypePostgres {
-		dbHandler.ViewRefreshHandler.RegisterViewRefreshHandler(_database.RefreshMaterializedViews)
-		go dbHandler.RefreshMaterializedViews()
-	}
+	dbHandler.ViewRefreshHandler.RegisterViewRefreshHandlers(config.DatabaseDriver)
+	go dbHandler.RefreshMaterializedViews()
 
 	if config.EnableFakeData {
 		go dbHandler.CreateFakeData()
--- a/backend/pkg/backend/backend.go
+++ b/backend/pkg/backend/backend.go
@@ -84,14 +84,14 @@ func Run() {
 	dbConfig := createDatabaseConfig(config)
 	dbHandler := _database.Init(dbConfig)
 
-	if config.EnableFakeData {
-		go dbHandler.CreateFakeData()
-	}
-
 	dbHandler.SetMaterializedViewHandler(dbConfig)
 	if config.DatabaseDriver == _database.DBDriverTypePostgres {
 		dbHandler.ViewRefreshHandler.RegisterViewRefreshHandler(_database.RefreshMaterializedViews)
 		go dbHandler.RefreshMaterializedViews()
+	}
+
+	if config.EnableFakeData {
+		go dbHandler.CreateFakeData()
 	}
 
 	backend := CreateBackend(dbHandler)
--- a/backend/pkg/backend/backend.go
+++ b/backend/pkg/backend/backend.go
@@ -84,8 +84,6 @@ func Run() {
 	dbConfig := createDatabaseConfig(config)
 	dbHandler := _database.Init(dbConfig)
 
-	dbHandler.SetMaterializedViewHandler(dbConfig)
-	dbHandler.ViewRefreshHandler.RegisterViewRefreshHandlers(config.DatabaseDriver)
 	go dbHandler.RefreshMaterializedViews()
 
 	if config.EnableFakeData {
--- a/backend/pkg/backend/backend.go
+++ b/backend/pkg/backend/backend.go
@@ -49,13 +49,14 @@ func CreateBackend(dbHandler *_database.Handler) *Backend {
 
 func createDatabaseConfig(config *_config.Config) *_database.DBConfig {
 	return &_database.DBConfig{
-		DriverType:     config.DatabaseDriver,
-		EnableInfoLogs: config.EnableDBInfoLogs,
-		DBPassword:     config.DBPassword,
-		DBUser:         config.DBUser,
-		DBHost:         config.DBHost,
-		DBPort:         config.DBPort,
-		DBName:         config.DBName,
+		DriverType:                config.DatabaseDriver,
+		EnableInfoLogs:            config.EnableDBInfoLogs,
+		DBPassword:                config.DBPassword,
+		DBUser:                    config.DBUser,
+		DBHost:                    config.DBHost,
+		DBPort:                    config.DBPort,
+		DBName:                    config.DBName,
+		ViewRefreshIntervalSecond: config.ViewRefreshIntervalSecond,
 	}
 }
 
@@ -87,7 +88,7 @@ func Run() {
 		go dbHandler.CreateFakeData()
 	}
 
-	dbHandler.SetMaterializedViewHandler()
+	dbHandler.SetMaterializedViewHandler(dbConfig)
 	if config.DatabaseDriver == _database.DBDriverTypePostgres {
 		dbHandler.ViewRefreshHandler.RegisterViewRefreshHandler(_database.RefreshMaterializedViews)
 		go dbHandler.RefreshMaterializedViews()
--- a/backend/pkg/config/config.go
+++ b/backend/pkg/config/config.go
@@ -34,7 +34,7 @@ const (
 	DBPortEnvVar              = "DB_PORT_NUMBER"
 	DatabaseDriver            = "DATABASE_DRIVER"
 	EnableDBInfoLogs          = "ENABLE_DB_INFO_LOGS"
-	ViewRefreshIntervalEnvVar = "VIEW_REFRESH_INTERVAL"
+	ViewRefreshIntervalEnvVar = "DB_VIEW_REFRESH_INTERVAL"
 
 	FakeDataEnvVar           = "FAKE_DATA"
 	FakeRuntimeScannerEnvVar = "FAKE_RUNTIME_SCANNER"
--- a/backend/pkg/config/config.go
+++ b/backend/pkg/config/config.go
@@ -27,13 +27,14 @@ const (
 	BackendRestPort    = "BACKEND_REST_PORT"
 	HealthCheckAddress = "HEALTH_CHECK_ADDRESS"
 
-	DBNameEnvVar     = "DB_NAME"
-	DBUserEnvVar     = "DB_USER"
-	DBPasswordEnvVar = "DB_PASS"
-	DBHostEnvVar     = "DB_HOST"
-	DBPortEnvVar     = "DB_PORT_NUMBER"
-	DatabaseDriver   = "DATABASE_DRIVER"
-	EnableDBInfoLogs = "ENABLE_DB_INFO_LOGS"
+	DBNameEnvVar              = "DB_NAME"
+	DBUserEnvVar              = "DB_USER"
+	DBPasswordEnvVar          = "DB_PASS"
+	DBHostEnvVar              = "DB_HOST"
+	DBPortEnvVar              = "DB_PORT_NUMBER"
+	DatabaseDriver            = "DATABASE_DRIVER"
+	EnableDBInfoLogs          = "ENABLE_DB_INFO_LOGS"
+	ViewRefreshIntervalEnvVar = "VIEW_REFRESH_INTERVAL"
 
 	FakeDataEnvVar           = "FAKE_DATA"
 	FakeRuntimeScannerEnvVar = "FAKE_RUNTIME_SCANNER"
@@ -44,15 +45,16 @@ type Config struct {
 	HealthCheckAddress string
 
 	// database config
-	DatabaseDriver           string
-	DBName                   string
-	DBUser                   string
-	DBPassword               string
-	DBHost                   string
-	DBPort                   string
-	EnableDBInfoLogs         bool
-	EnableFakeData           bool
-	EnableFakeRuntimeScanner bool
+	DatabaseDriver            string
+	DBName                    string
+	DBUser                    string
+	DBPassword                string
+	DBHost                    string
+	DBPort                    string
+	EnableDBInfoLogs          bool
+	EnableFakeData            bool
+	EnableFakeRuntimeScanner  bool
+	ViewRefreshIntervalSecond int
 }
 
 func LoadConfig() (*Config, error) {
@@ -67,6 +69,7 @@ func LoadConfig() (*Config, error) {
 	config.DBHost = viper.GetString(DBHostEnvVar)
 	config.DBPort = viper.GetString(DBPortEnvVar)
 	config.DBName = viper.GetString(DBNameEnvVar)
+	config.ViewRefreshIntervalSecond = viper.GetInt(ViewRefreshIntervalEnvVar)
 	config.EnableDBInfoLogs = viper.GetBool(EnableDBInfoLogs)
 	config.EnableFakeData = viper.GetBool(FakeDataEnvVar)
 	config.EnableFakeRuntimeScanner = viper.GetBool(FakeRuntimeScannerEnvVar)
--- a/backend/pkg/database/application.go
+++ b/backend/pkg/database/application.go
@@ -86,6 +86,7 @@ type ApplicationTableHandler struct {
 	licensesView      *gorm.DB
 	IDsView           IDsView
 	db                *gorm.DB
+	driverType        string
 }
 
 func (Application) TableName() string {
@@ -120,6 +121,8 @@ func (a *ApplicationTableHandler) Create(app *Application, params *TransactionPa
 		Create(app).Error; err != nil {
 		return fmt.Errorf("failed to create application: %v", err)
 	}
+
+	refreshMaterializedViewsIfNeeded(a.db, a.driverType)
 
 	return nil
 }
--- a/backend/pkg/database/application.go
+++ b/backend/pkg/database/application.go
@@ -122,7 +122,7 @@ func (a *ApplicationTableHandler) Create(app *Application, params *TransactionPa
 		return fmt.Errorf("failed to create application: %v", err)
 	}
 
-	a.setViewRefreshHandlerIfExists()
+	a.tableChanged()
 
 	return nil
 }
@@ -133,7 +133,7 @@ func (a *ApplicationTableHandler) UpdateInfo(app *Application, params *Transacti
 		return fmt.Errorf("failed to update application info: %v", err)
 	}
 
-	a.setViewRefreshHandlerIfExists()
+	a.tableChanged()
 
 	return nil
 }
@@ -151,7 +151,7 @@ func (a *ApplicationTableHandler) Delete(app *Application) error {
 		return fmt.Errorf("failed to delete application: %v", err)
 	}
 
-	a.setViewRefreshHandlerIfExists()
+	a.tableChanged()
 
 	return nil
 }
@@ -411,8 +411,8 @@ func (a *ApplicationTableHandler) setCountFilters(tx *gorm.DB, filters *CountFil
 	return tx
 }
 
-func (a *ApplicationTableHandler) setViewRefreshHandlerIfExists() {
+func (a *ApplicationTableHandler) tableChanged() {
 	if a.viewRefreshHandler != nil {
-		a.viewRefreshHandler.SetTrue()
+		a.viewRefreshHandler.TableChanged(applicationTableName)
 	}
 }
--- a/backend/pkg/database/application.go
+++ b/backend/pkg/database/application.go
@@ -81,12 +81,12 @@ type ApplicationTable interface {
 }
 
 type ApplicationTableHandler struct {
-	applicationsTable *gorm.DB
-	applicationsView  *gorm.DB
-	licensesView      *gorm.DB
-	IDsView           IDsView
-	db                *gorm.DB
-	driverType        string
+	applicationsTable  *gorm.DB
+	applicationsView   *gorm.DB
+	licensesView       *gorm.DB
+	IDsView            IDsView
+	db                 *gorm.DB
+	viewRefreshHandler *ViewRefreshHandler
 }
 
 func (Application) TableName() string {
@@ -122,7 +122,7 @@ func (a *ApplicationTableHandler) Create(app *Application, params *TransactionPa
 		return fmt.Errorf("failed to create application: %v", err)
 	}
 
-	refreshMaterializedViewsIfNeeded(a.db, a.driverType)
+	a.setViewRefreshHandlerIfExists()
 
 	return nil
 }
@@ -133,7 +133,7 @@ func (a *ApplicationTableHandler) UpdateInfo(app *Application, params *Transacti
 		return fmt.Errorf("failed to update application info: %v", err)
 	}
 
-	refreshMaterializedViewsIfNeeded(a.db, a.driverType)
+	a.setViewRefreshHandlerIfExists()
 
 	return nil
 }
@@ -151,7 +151,7 @@ func (a *ApplicationTableHandler) Delete(app *Application) error {
 		return fmt.Errorf("failed to delete application: %v", err)
 	}
 
-	refreshMaterializedViewsIfNeeded(a.db, a.driverType)
+	a.setViewRefreshHandlerIfExists()
 
 	return nil
 }
@@ -409,4 +409,10 @@ func (a *ApplicationTableHandler) setCountFilters(tx *gorm.DB, filters *CountFil
 	tx = CISDockerBenchmarkLevelFilterGte(tx, columnCISDockerBenchmarkLevelCountersHighestLevel, filters.CisDockerBenchmarkLevelGte)
 
 	return tx
+}
+
+func (a *ApplicationTableHandler) setViewRefreshHandlerIfExists() {
+	if a.viewRefreshHandler != nil {
+		a.viewRefreshHandler.SetTrue()
+	}
 }
--- a/backend/pkg/database/application.go
+++ b/backend/pkg/database/application.go
@@ -133,6 +133,8 @@ func (a *ApplicationTableHandler) UpdateInfo(app *Application, params *Transacti
 		return fmt.Errorf("failed to update application info: %v", err)
 	}
 
+	refreshMaterializedViewsIfNeeded(a.db, a.driverType)
+
 	return nil
 }
 
@@ -148,6 +150,8 @@ func (a *ApplicationTableHandler) Delete(app *Application) error {
 	if err := a.db.Delete(app).Error; err != nil {
 		return fmt.Errorf("failed to delete application: %v", err)
 	}
+
+	refreshMaterializedViewsIfNeeded(a.db, a.driverType)
 
 	return nil
 }
--- a/backend/pkg/database/application.go
+++ b/backend/pkg/database/application.go
@@ -122,7 +122,7 @@ func (a *ApplicationTableHandler) Create(app *Application, params *TransactionPa
 		return fmt.Errorf("failed to create application: %v", err)
 	}
 
-	a.tableChanged()
+	a.viewRefreshHandler.TableChanged(applicationTableName)
 
 	return nil
 }
@@ -133,7 +133,7 @@ func (a *ApplicationTableHandler) UpdateInfo(app *Application, params *Transacti
 		return fmt.Errorf("failed to update application info: %v", err)
 	}
 
-	a.tableChanged()
+	a.viewRefreshHandler.TableChanged(applicationTableName)
 
 	return nil
 }
@@ -151,7 +151,7 @@ func (a *ApplicationTableHandler) Delete(app *Application) error {
 		return fmt.Errorf("failed to delete application: %v", err)
 	}
 
-	a.tableChanged()
+	a.viewRefreshHandler.TableChanged(applicationTableName)
 
 	return nil
 }
@@ -409,10 +409,4 @@ func (a *ApplicationTableHandler) setCountFilters(tx *gorm.DB, filters *CountFil
 	tx = CISDockerBenchmarkLevelFilterGte(tx, columnCISDockerBenchmarkLevelCountersHighestLevel, filters.CisDockerBenchmarkLevelGte)
 
 	return tx
-}
-
-func (a *ApplicationTableHandler) tableChanged() {
-	if a.viewRefreshHandler != nil {
-		a.viewRefreshHandler.TableChanged(applicationTableName)
-	}
 }
--- a/backend/pkg/database/database.go
+++ b/backend/pkg/database/database.go
@@ -39,25 +39,25 @@ const (
 // In the case of materialized views slow queries can happen when refreshing the views,
 // but reading these views will not cause slow queries when navigating on the UI.
 // Materialized view is only supported by PostgreSQL.
-const (
-	packagesView        = "packages_view"
-	resourcesView       = "resources_view"
-	applicationsView    = "applications_view"
-	vulnerabilitiesView = "vulnerabilities_view"
-)
+var materializedViews = []string{
+	applicationViewName,
+	resourceViewName,
+	packageViewName,
+	vulnerabilityViewName,
+}
 
 // order is important, need to drop views in a reverse order of the creation.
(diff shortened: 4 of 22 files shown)


backend/pkg/database/application.go
```diff
@@ -121,6 +122,8 @@ func (a *ApplicationTableHandler) Create(app *Application, params *TransactionPa
 		return fmt.Errorf("failed to create application: %v", err)
 	}
 
+	refreshMaterializedViewsIfNeeded(a.db, a.driverType)
+
```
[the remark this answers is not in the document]
sambetts-cisco: it also seems like this might need some sort of mapping to know which MaterializedViews should be updated based on what has changed. If we expand the views which are materialized in the future it would be weird/inefficent to refresh the CIS benchmark view when the packages DB had new data.

backend/pkg/database/scheduler.go
```diff
@@ -72,6 +74,8 @@ func (s *SchedulerTableHandler) Set(scheduler *Scheduler) error {
 		return fmt.Errorf("failed to set scheduler: %v", err)
 	}
 
+	refreshMaterializedViewsIfNeeded(s.db, s.driverType)
+
 	return nil
```
sambetts-cisco, line 79: Data is updated in the "UpdateNextScanTime" function below should we have a view refresh there too?

backend/pkg/database/vulnerability.go
```diff
@@ -137,6 +139,11 @@ func (v *VulnerabilityTableHandler) Create(vul *Vulnerability) error {
 	if err := v.vulnerabilitiesTable.Create(vul).Error; err != nil {
 		return fmt.Errorf("failed to create vulnerability: %v", err)
 	}
+
+	if v.driverType == DBDriverTypePostgres {
+		refreshMaterializedViewsIfNeeded(v.db, v.driverType)
+	}
```
sambetts-cisco, line 145: no need for the wrapper "if v.driverType" here, its handled inside of refreshMaterializedViewsIfNeeded.

sambetts-cisco (changes requested):
Hey Peter, got a few comments that need addressing, and a couple of questions too. Can you also provide more information in the commit message about why we choose materialised views, and why its only for the application, resources and packages views.

fishkerez:
@pbalogh-sa did considered also using indexes on the materialized views? it seems like it is possible:
https://spin.atomicobject.com/2018/04/09/postgres-materialized-views/

backend/pkg/database/application.go
```diff
@@ -86,6 +86,7 @@ type ApplicationTableHandler struct {
 	licensesView      *gorm.DB
 	IDsView           IDsView
 	db                *gorm.DB
+	driverType        string
```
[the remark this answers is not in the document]
pbalogh-sa: dbDriverType is not used anymore here

backend/pkg/database/application.go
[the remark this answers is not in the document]
pbalogh-sa: updated

backend/pkg/database/database.go
```diff
@@ -568,14 +604,23 @@ func createAllViews(db *gorm.DB) {
...
 	}
 }
 
-func dropViewIfExists(db *gorm.DB, viewName string) {
-	if err := db.Exec(fmt.Sprintf("DROP VIEW IF EXISTS %s", viewName)).Error; err != nil {
+func dropViewIfExists(db *gorm.DB, viewName, dropCommand string) {
+	if err := db.Exec(fmt.Sprintf(dropCommand, viewName)).Error; err != nil {
 		log.Fatalf("Failed to drop %s: %v", viewName, err)
 	}
 }
```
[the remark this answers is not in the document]
pbalogh-sa: You are right, it's better to decide the `dropCommand` in `dropViewIfExists` function

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,81 @@
...
+type ViewRefreshHandler struct {
+	mu                 sync.Mutex
+	shouldRefreshViews bool
+}
+
+func (vh *ViewRefreshHandler) SetTrue() {
+	vh.mu.Lock()
+	defer vh.mu.Unlock()
+	vh.shouldRefreshViews = true
+}
```
sambetts-cisco, line 38: If we let this function accept a string and we added `tablesChanged map[string]bool` to the ViewRefreshHandler we could create a set of which tables have changed, with that information we could change the view handler to allow views to register their own refresh handlers based on different table changes. I'm thinking something like:

```
type refreshfunc func(db *gorm.DB)

type ViewRefreshHandler struct {
	mu                        sync.Mutex
	tablesChanged   map[string]bool
	refreshHandlers map[string][]refreshfunc
}

func (vh *ViewRefreshHandler) TableChanged(table string) {
    	vh.mu.Lock()
	defer vh.mu.Unlock()
	vh.tablesChanged[table] = true
}

func (vh *ViewRefreshHandler) GetAndClearChanges() map[string]bool {
	vh.mu.Lock()
	defer vh.mu.Unlock()
	changes := vh.tableChanged
	vh.tableChanged = map[string]bool{}
	return changes
}

func (vh *ViewRefreshHandler) RegisterViewRefreshHandler(f refreshfunc, tables ...string) {
        for _, table := range tables {
            vh.refreshHandlers[table] = append(vh.RefreshHandlers[table], f)
        }
}

func (vh *ViewRefreshHandler) DoRequiredRefreshes(db *gorm.DB) {
	changedTables = vh.GetAndClearChanges()
         for table, _ := range changedTables {
              handlers := vh.refreshHandlers[table]
              for _, handler := range handlers {
                  handler(db)
              }
         }
}

func (db *Handler) RefreshMaterializedViews() {
	ticker := time.NewTicker(materializedViewRefreshIntervalSecond * time.Second)
	
	done := make(chan bool)
	for {
		select {
		case <-done:
			return
		case <-ticker.C:
                        db.ViewRefreshHandler.DoRequiredRefreshes(db)
		}
	}
}
```

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (db *Handler) RefreshMaterializedViews() {
+	ticker := time.NewTicker(materializedViewRefreshIntervalSecond * time.Second)
+	done := make(chan bool)
+
+	for {
+		select {
+		case <-done:
+			return
+		case <-ticker.C:
```
sambetts-cisco, line 60: One issue with using a ticker here is that if the view refresh takes longer than materializedViewRefreshIntervalSecond it will instantly run the refresh again after last refresh, this could mean that when the views get very large we are constantly refreshing all the views. Perhaps we should change this to `<-time.After(...)` so that the timer will only begin again after we fully complete the previous refresh?

backend/pkg/database/resource.go
```diff
@@ -520,3 +524,9 @@ func (r *ResourceTableHandler) setCountFilters(tx *gorm.DB, filters *CountFilter
 
 	return tx, nil
 }
+
+func (r *ResourceTableHandler) setViewRefreshHandlerIfExists() {
+	if r.viewRefreshHandler != nil {
+		r.viewRefreshHandler.SetTrue()
+	}
+}
```
sambetts-cisco, line 532: if we change the code so that the views can register refresh handlers at the point they are created, then we can always have a viewRefreshHandler, and have createAllViews only register refresh funcs if DB type == postgres, or in the future maybe some other DB type?

backend/pkg/database/scheduler.go
sambetts-cisco: we might still need a s.setViewRefreshHandlerIfExists() in the UpdateNextScanTime below

sambetts-cisco (changes requested):
Overall looks good, I think there may be one missing `o.setViewRefreshHandlerIfExists()` in an update function. I've also got a suggestion that may make this a little more flexible for the future if we add new views and also it could help reduce the number of views that have to be rebuilt on each tick.

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+)
+
+const materializedViewRefreshIntervalSecond = 5
```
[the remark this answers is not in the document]
sambetts-cisco: +1 configurable would be a good idea so that we can tweak it without needing to rebuild the world

backend/pkg/database/refresh_materialized_views.go
pbalogh-sa: Thanks for the suggestion, I will do it this way.

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,81 @@
...
+		log.Fatalf("Failed to refresh materialized %s: %v", applicationsView, err)
+	}
+	if err := db.Exec(fmt.Sprintf(refreshMaterializedViewCommand, resourcesView)).Error; err != nil {
+		log.Fatalf("Failed to refresh materialized %s: %v", resourcesView, err)
+	}
+	if err := db.Exec(fmt.Sprintf(refreshMaterializedViewCommand, packagesView)).Error; err != nil {
+		log.Fatalf("Failed to refresh materialized %s: %v", packagesView, err)
+	}
+	if err := db.Exec(fmt.Sprintf(refreshMaterializedViewCommand, vulnerabilitiesView)).Error; err != nil {
+		log.Fatalf("Failed to refresh materialized %s: %v", vulnerabilitiesView, err)
```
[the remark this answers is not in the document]
pbalogh-sa: No, I will update.

backend/pkg/database/refresh_materialized_views.go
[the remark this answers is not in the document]
pbalogh-sa: Yes, it would be better.

e2e/cli_scan_test.go
```diff
@@ -68,16 +69,22 @@ func TestCLIScan(t *testing.T) {
 			// analyze image with --merge-sbom directory sbom, and export to backend
 			t.Logf("analyze image...")
 			analyzeImage(t, DirectoryAnalyzeOutputSBOMFile, appID)
+			// wait for refreshing materialized views
+			time.Sleep(common.WaitForMaterializedViewRefreshSecond * time.Second)
```
fishkerez, line 73: I think it is better to check the db every 1 second in order to see if the the refresh has ended, I dont like entering a sleep to the tests

pbalogh-sa: How can we decide if the refresh is not ended or the refresh is ended but the results are not the expected?

backend/pkg/config/config.go
```diff
@@ -27,13 +27,14 @@ const (
...
-	DatabaseDriver   = "DATABASE_DRIVER"
-	EnableDBInfoLogs = "ENABLE_DB_INFO_LOGS"
+	DBNameEnvVar              = "DB_NAME"
+	DBUserEnvVar              = "DB_USER"
+	DBPasswordEnvVar          = "DB_PASS"
+	DBHostEnvVar              = "DB_HOST"
+	DBPortEnvVar              = "DB_PORT_NUMBER"
+	DatabaseDriver            = "DATABASE_DRIVER"
+	EnableDBInfoLogs          = "ENABLE_DB_INFO_LOGS"
+	ViewRefreshIntervalEnvVar = "VIEW_REFRESH_INTERVAL"
```
akpsgit, line 37: Maybe to call the env var DB_VIEW_REFRESH_INTERVAL

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,125 @@
...
+		if err := db.Exec(fmt.Sprintf(refreshMaterializedViewCommand, viewName)).Error; err != nil {
+			log.Errorf("Failed to refresh materialized %s: %v", viewName, err)
+		}
+	}
+}
+
+func initMaterializedViews(db *gorm.DB, viewNames []string) {
+	for _, viewName := range viewNames {
+		if err := db.Exec(fmt.Sprintf(initMaterializedViewCommand, viewName)).Error; err != nil {
+			log.Errorf("Failed to refresh materialized %s: %v", viewName, err)
```
FrimIdan, line 122: Failed to **init**?

backend/pkg/backend/backend.go
```diff
@@ -87,6 +88,12 @@ func Run() {
 		go dbHandler.CreateFakeData()
 	}
 
+	dbHandler.SetMaterializedViewHandler(dbConfig)
```
fishkerez, line 91: Do we want to set it in case of local DB? (sqlite) I dont see it is being used in that case

backend/pkg/database/database.go
```diff
@@ -472,14 +518,16 @@ func initDataBase(config *DBConfig) *gorm.DB {
 
 	setupJoinTables(db)
 
+	// drop views before auto migrate
```
akpsgit, line 521: this is probably conflict with https://github.com/openclarity/kubeclarity/commit/dbec182096b287131bba6a61331bc88d17070e8e, need to rebase

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,125 @@
...
+	"gorm.io/gorm"
+)
+
+const DefaultViewRefreshIntervalSecond = 5
+
+type refreshFunc func(db *gorm.DB, viewNames []string)
+
+type ViewRefreshHandler struct {
+	mu                        sync.Mutex
+	viewsToRefresh            map[string][]string
```
FrimIdan, line 33: please add a comment what is mapping to what

backend/pkg/database/database.go
```diff
@@ -35,24 +35,44 @@ const (
...
 	"licenses_view",
 }
 
+const (
+	createViewCommand              = "CREATE VIEW"
+	createMaterializedViewCommand  = "CREATE MATERIALIZED VIEW"
+	dropViewCommand                = "DROP VIEW IF EXISTS %s"
+	dropMaterializedViewCommand    = "DROP MATERIALIZED VIEW IF EXISTS %s;"
+	refreshMaterializedViewCommand = "REFRESH MATERIALIZED VIEW CONCURRENTLY %s;"
+	initMaterializedViewCommand    = "REFRESH MATERIALIZED VIEW %s;"
```
fishkerez, line 73: init? or refresh?

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,125 @@
...
+	vh.refreshFunc(db, viewNames)
+}
+
+func (vh *ViewRefreshHandler) getViewsToRefresh() []string {
+	viewToRefresh := make(map[string]bool)
+	tables := vh.GetAndClearChanges()
+	for table, ok := range tables {
+		if ok {
+			for _, viewName := range vh.viewsToRefresh[table] {
+				viewToRefresh[viewName] = true
```
FrimIdan, line 64: why there is a need to use `viewToRefresh` map and not creating the `views` slice here?

If this is done in order to remove duplication a comment here will be nice.

backend/pkg/database/database.go
```diff
@@ -543,39 +599,70 @@ func createAllViews(db *gorm.DB) {
...
 		log.Fatalf("Failed to create vulnerabilities_view: %v", err)
 	}
 
 	if err := db.Exec(cisDockerBenchmarkResultsViewQuery).Error; err != nil {
 		log.Fatalf("Failed to create cis_d_b_checks_view: %v", err)
 	}
+
+	if dbDriver == DBDriverTypePostgres {
+		createIndexForMaterializedViews(db)
+		initMaterializedViews(db, materializedViews)
```
FrimIdan, line 628: shouldn't we use `log.Fatalf` here if `initMaterializedViews` failed?

backend/pkg/database/database.go
```diff
@@ -318,6 +350,10 @@ GROUP BY vulnerabilities.id,
          pv.fix_version;
   `
 
+	createVulnerabilitiesViewIndex = `
+CREATE UNIQUE INDEX vulnerabilities_view_id ON vulnerabilities_view (id, package_id);
```
FrimIdan, line 354: why here the index is on `(id, package_id)` and not only `id` as all other?

backend/pkg/database/new_vulnerability.go
```diff
@@ -133,5 +137,13 @@ func (v *NewVulnerabilityTableHandler) CreateNewVulnerabilitiesTrends(applicatio
...
 	}
 
+	v.tableChanged()
+
 	return nil
 }
+
+func (v *NewVulnerabilityTableHandler) tableChanged() {
+	if v.viewRefreshHandler != nil {
+		v.viewRefreshHandler.TableChanged("vulnerabilities")
```
fishkerez, line 147: please use the const for "vulnerabilities"

e2e/cli_scan_test.go
fishkerez: Is there is no indication of refresh ended?

backend/pkg/database/new_vulnerability.go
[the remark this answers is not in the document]
akpsgit: sorry, new_vulnerabilities not related to materialized views. but does creating the trends impact the vulnerabilities table?

backend/pkg/database/database.go
pbalogh-sa: We can refresh materialized views concurrently, but it isn't working without populated data, so this is the reason why I used `init` in the const name.

e2e/cli_scan_test.go
pbalogh-sa: From the cli we cannot check the end of view refresh.

backend/pkg/database/database.go
fishkerez: I think it is better to call it refreshMaterializedViewCommand and refreshMaterializedViewConcurrentlyCommand just for the clarity of things... what ever you think

backend/pkg/database/new_vulnerability.go
pbalogh-sa: ok

backend/pkg/database/database.go
pbalogh-sa: Refreshing conrurrently a view requires a unique index, but if we use only the id for the created index it will seem duplicated in the case of vulnerabilities.
```
ERROR: duplicate key value violates unique constraint "vulnerabilities_view_id" (SQLSTATE 23505)
```

backend/pkg/database/new_vulnerability.go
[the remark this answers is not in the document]
pbalogh-sa: Not needed here, you are right. I removed it.

backend/pkg/database/refresh_materialized_views.go
sambetts-cisco: @FrimIdan the viewToRefresh map here is actually being used as a set, if two tables changed which require the same view then we don't want to post-process a slice to remove the duplicates for the next step

backend/pkg/backend/backend.go
```diff
@@ -83,6 +84,12 @@ func Run() {
 	dbConfig := createDatabaseConfig(config)
 	dbHandler := _database.Init(dbConfig)
 
+	dbHandler.SetMaterializedViewHandler(dbConfig)
+	if config.DatabaseDriver == _database.DBDriverTypePostgres {
+		dbHandler.ViewRefreshHandler.RegisterViewRefreshHandler(_database.RefreshMaterializedViews)
```
sambetts-cisco, line 89: I think perhaps you misunderstood the intension of RegistryViewRefreshHandler in my previous comment, the intension what that inside the of the createAllViews function you would use it to register a refresh handler for each view for example:
```
func createAllViews(db *gorm.DB, dbDriver string) {
        ....

	if err := db.Exec(fmt.Sprintf(applicationsViewQuery, createCommand)).Error; err != nil {
		log.Fatalf("Failed to create applications_view: %v", err)
	}
	if dbDriver == DBDriverTypePostgres {
	       // Register postgres view refresh *if* applications, resources or packages tables change
	       // those are the only tables that view requires.
	       dbHandler.ViewRefreshHandler.RegisterViewRefreshHandler(
	               func(db *gorm.DB){
	                       if err := db.Exec(fmt.Sprintf(refreshMaterializedViewConcurrentlyCommand, applicationsViewName)).Error; err != nil {
	                               log.Errorf("Failed to refresh materialized %s: %v", applicationsViewName, err)
		               }
                       },
                       "applications",
                       "resources",
                       "packages",
               )
	}
	
	...
	
        if err := db.Exec(fmt.Sprintf(vulnerabilitiesViewQuery, createCommand)).Error; err != nil {
		log.Fatalf("Failed to create vulnerabilities_view: %v", err)
	}
	if dbDriver == DBDriverTypePostgres {
	       // Register postgres view refresh *if* vulnerability, resources or packages tables change
	       dbHandler.ViewRefreshHandler.RegisterViewRefreshHandler(
	               func(db *gorm.DB){
	                       if err := db.Exec(fmt.Sprintf(refreshMaterializedViewConcurrentlyCommand, vulnerabilityViewName)).Error; err != nil {
	                               log.Errorf("Failed to refresh materialized %s: %v", vulnerabilityViewName, err)
		               }
                       },
                       "vulnerabilities",
                       "resources",
                       "packages",
               )
        }
	...
}
```

backend/pkg/database/new_vulnerability.go
```diff
@@ -68,6 +68,7 @@ func (v *NewVulnerabilityTableHandler) Create(vul *NewVulnerability) error {
 	if err := v.newVulnerabilitiesTable.Create(vul).Error; err != nil {
 		return fmt.Errorf("failed to create new vulnerability: %v", err)
 	}
+
```
sambetts-cisco, line 71: nit: random extra whitespace change

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+		}
+	}
+	views := make([]string, 0)
+	for viewName, ok := range viewToRefresh {
+		if ok {
+			views = append(views, viewName)
+		}
+	}
+	return views
```
sambetts-cisco, line 74: honestly I wouldn't even convert this to a list of strings when you can just loop over the set like in my previous review

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+func initMaterializedViews(db *gorm.DB, viewNames []string) {
+	for _, viewName := range viewNames {
+		if err := db.Exec(fmt.Sprintf(refreshMaterializedViewCommand, viewName)).Error; err != nil {
+			log.Fatalf("Failed to init materialized %s: %v", viewName, err)
+		}
+	}
+}
```
sambetts-cisco, line 129: We could remove this function and instead have the createAllViews function in the DB just mark the tables as changed, i.e.
```
func createAllViews(...) {
        ....
         
	if err := db.Exec(fmt.Sprintf(applicationsViewQuery, createCommand)).Error; err != nil {
		log.Fatalf("Failed to create applications_view: %v", err)
	}
	if dbDriver == DBDriverTypePostgres {
	       // Register postgres view refresh *if* applications, resources or packages tables change
	       // those are the only tables that view requires.
	       dbHandler.ViewRefreshHandler.RegisterViewRefreshHandler(
	               func(db *gorm.DB){
	                       if err := db.Exec(fmt.Sprintf(refreshMaterializedViewConcurrentlyCommand, applicationsViewName)).Error; err != nil {
	                               log.Errorf("Failed to refresh materialized %s: %v", applicationsViewName, err)
		               }
                       },
                       "applications",
                       "resources",
                       "packages",
               )
               // Mark the tables the view is interested in as changed so that the view refresh
               // handler will refresh this view at least once on the next cycle even though the
               // tables have not really changed
               dbHandler.ViewRefreshHandler.TableChanged("applications")
               dbHandler.ViewRefreshHandler.TableChanged("resources")
               dbHandler.ViewRefreshHandler.TableChanged("packages")
	}
	
	....
}
```

e2e/runtime_scan_test.go
```diff
@@ -50,6 +51,8 @@ func TestRuntimeScan(t *testing.T) {
 			assert.NilError(t, waitForScanDone(t))
 
 			t.Logf("get runtime scan results...")
+			// wait for refreshing materialized views
+			time.Sleep(common.WaitForMaterializedViewRefreshSecond * time.Second)
```
sambetts-cisco, line 55: in the tests we could configure the viewRefreshHandler time significantly less than default like 1 second, that way we're not waiting for so long in the tests for things to occur.

sambetts-cisco (changes requested):

backend/pkg/database/refresh_materialized_views.go
FrimIdan: Yep, so a comment in the code will be nice IMO

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,129 @@
...
+	for table, ok := range tables {
+		if ok {
+			for _, viewName := range vh.viewsToRefresh[table] {
+				viewToRefresh[viewName] = true
+			}
+		}
+	}
+	views := make([]string, 0)
+	for viewName, ok := range viewToRefresh {
+		if ok {
```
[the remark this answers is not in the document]
FrimIdan: we can always use map[]struct{}.  (:

https://itnext.io/set-in-go-map-bool-and-map-struct-performance-comparison-5315b4b107b
> From the result above, map[]struct{} is 5% faster in time and 10% less memory consumption comparing to map[]bool when it comes to a big Set.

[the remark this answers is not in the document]
sambetts-cisco: damn thats super interesting, the internals of go must compile struct{} down to literally nothing

backend/pkg/database/refresh_materialized_views.go
pbalogh-sa: In the case of all related tables not being populated, refreshing the view concurrently will fail. So this is the reason for the first time we need to refresh it without the `CONCURRENTLY` option.

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,129 @@
...
+		resourceTableName,
+		packageTableName,
+		vulnerabilityTableName,
+	}
+	for _, table := range tables {
+		viewsToRefresh[table] = materializedViews
+	}
+
+	return viewsToRefresh
+}
```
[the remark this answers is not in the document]
pbalogh-sa: In our case, we need to refresh all views on tables changed, for example, if the application table is updated we need to refresh the vulnerabilities view because the vulnerabilities view contains fields from packages, applications, and resources table.

backend/pkg/backend/backend.go
pbalogh-sa: I understood what you wanted in the previous review, but I think we don't need to register a handler for all views because we want to refresh only one way in case of a specified dbDriver. So we need to register the handler for a dbDriver and do the refresh iterating over the required views.

sambetts-cisco: that won't be as generic and flexible as registering a refresh handler per view, it might satisfy our current use case but it doesn't leave much room for expansion/extension in the future if we need to do more things for one view over a different view. If you want to reduce duplication but keep the flexibility then you can just do:

```
func createAllViews() {
       ...
       dbHandler.ViewRefreshHandler.RegisterViewRefreshHandler(
            getPostgresViewRefreshHandlerFunc(vulnerabilityViewName),
            "table1",
            "table2",
            "table3",
            "table4",
       )
       ...
}


func getPostgresViewRefreshHandlerFunc(viewname string) func(db *gorm.DB) {
     return func(db *gorm.DB) {
             if err := db.Exec(fmt.Sprintf(refreshMaterializedViewConcurrentlyCommand, viewname)).Error; err != nil {
	             log.Errorf("Failed to refresh materialized %s: %v", viewname, err)
             }
     }
}
```

this still leaves the option for a view that needs to it have its own handler that does other things.

backend/pkg/database/refresh_materialized_views.go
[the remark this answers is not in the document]
pbalogh-sa: I removed the check. I think we can use map[string]bool because it can't be a huge set.

backend/pkg/database/refresh_materialized_views.go
pbalogh-sa: We iterate over the map of tables that were changed and based on the mapping between tables and views in this way we create a simple list that can be passed to the `refreshFunc` to iterate over it.

sambetts-cisco: you already have a simple list (a set) viewsToRefresh, you can iterate over it like:

```
for viewName, _ := viewsToRefresh {
     refreshFunc(viewName)
}
```

no need to convert its just a waste of time and memory.

pbalogh-sa: Oh, you mean only the last conversion. You're right. I will udate

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,163 @@
...
+		vh.registerViewRefreshHandler(getPostgresViewRefreshHandlerFunc(vulnerabilityViewName),
+			vulnerabilityViewName,
+			[]string{
+				applicationTableName,
+				resourceTableName,
+				packageTableName,
+				vulnerabilityTableName,
+			})
+	}
+}
```
sambetts-cisco, line 113: This still doesn't put this close to the actual view creation and management, I would rather this was a function called alongside initPostgresMaterializedViews like "registerPostgresViewRefreshFunctions". Its not the responsiblity of the ViewRefreshHandler to know anything about the DB driver.

Also if you use `...string` in the function definition for registerViewRefreshHandler there is no need for the extra `[]string{...}` here.

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (vh *ViewRefreshHandler) registerViewRefreshHandler(f refreshFunc, viewName string, tables []string) {
+	vh.refreshFunc[viewName] = f
+	vh.addViewsToRefreshByTable(viewName, tables)
```
sambetts-cisco, line 117: not sure why this needed to be a second function I think this could just be inlined.

I like having two maps one for the callbacks and then one to map the callbacks by name to the tables as it will allow us to keep track of all the callbacks which have been registered.

However I think we should change the viewsToRefresh map to a `map[string]<set>` to ensure that a callback can only be added once to a particular table the current append @L126 will not check this.

changing viewsToRefresh too:

```
type viewNameSet map[string]struct{}

type ViewRefreshHandler struct {
    ...
    viewsToRefresh map[string]viewNameSet
    ...
}
```

and then:

```
func (vh *ViewRefreshHandler) addViewsToRefreshByTable(viewName string, tables []string) {
	for _, table := range tables {
                if _, ok := vh.viewsToRefresh[table]; ok {
                        vh.viewsToRefresh[table] = viewNameSet{}
                }
	        vh.viewsToRefresh[table][viewName] = struct{}{}
	}
}
```

will ensure a view can only be in a table's list to refresh once and its safe to call the register function multiple times. To make it even safer to re-register we could even clear all other tables using that view like:

```
...
        for _, views := range vh.viewsToRefresh {
                delete(views, viewName)
        }
	for _, table := range tables {
                if _, ok := vh.viewsToRefresh[table]; ok {
                        vh.viewsToRefresh[table] = viewNameSet{}
                }
	        vh.viewsToRefresh[table][viewName] = struct{}{}
	}
...
```

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,163 @@
...
+		viewsToRefresh:            make(map[string][]string),
+		viewRefreshIntervalSecond: time.Duration(config.ViewRefreshIntervalSecond) * time.Second,
+		refreshFunc:               map[string]refreshFunc{},
+	}
+}
+
+func (db *Handler) RefreshMaterializedViews() {
+	if !db.ViewRefreshHandler.IsSetViewRefreshHandler() {
+		return
+	}
```
sambetts-cisco, line 142: I don't think this should be needed, the view handler should just always be started and running in the background, whether it does anything should be entirely controlled by registering refresh handlers at the point we are creating the views.

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,163 @@
...
+	for {
+		<-time.After(db.ViewRefreshHandler.viewRefreshIntervalSecond)
```
sambetts-cisco, line 144: Don't we want a way to cleanly shut this down? a cancel channel or context? Maybe it doesn't matter too much

backend/pkg/database/refresh_materialized_views.go
```diff
@@ -0,0 +1,163 @@
...
+		db.ViewRefreshHandler.runRequiredRefreshes(db.DB)
+	}
+}
+
+func getPostgresViewRefreshHandlerFunc(viewName string) func(db *gorm.DB) {
+	return func(db *gorm.DB) {
+		if err := db.Exec(fmt.Sprintf(refreshMaterializedViewConcurrentlyCommand, viewName)).Error; err != nil {
```
sambetts-cisco, line 151: When I was talking about concurrency before I wasn't refering to this postgres's concurrency, when I talked about concurrency before I meant:

```
func (vh *ViewRefreshHandler) runRequiredRefreshes(db *gorm.DB) {
	viewNames := vh.getViewsToRefresh()
	var wg sync.WaitGroup
	for viewName := range viewNames {
	        wg.Add(1)
	        go func() {
	                defer wg.Done()
		        vh.refreshFunc[viewName](db)
		}()
	}
	wg.Wait()
}
```

unsure if gorm/postgres will handle us doing that, but it seems like we should need to serially run the Exec's and then wait on them to be done before calling the next one.

Postgres's CONCURRENTLY option doesn't actually refresh the view concurrently, it just doesn't lock the table during the refresh.

sambetts-cisco (changes requested):

backend/pkg/database/refresh_materialized_views.go
pbalogh-sa: Ok, done

sambetts-cisco (approved):
Functionally everything LGTM, I have a couple more comments for refactoring db handler and view refresh handler so they are not interlinked so heavily.
The PR is close enough, and is passing all the tests and fixes some real world problems, so I'm happy to approve it. I would like to see the refactor done in a follow up if possible, it'll make maintaining and unit testing the code much easier long term

akpsgit (approved):

sambetts-cisco (approved):

backend/pkg/database/refresh_materialized_views.go
pbalogh-sa: i've added comments

PR #241, opened 2022-10-27
Fix (sbomdb): convert application log level to gorm logger level

pbalogh-sa:
Signed-off-by: Peter Balogh <p.balogh.sa@gmail.com>

Use the application log level in the case of database logging as well.


pbalogh-sa:
Use separate log level for DB: #242

PR #252, opened 2022-11-07
Add support for Trivy as an SBOM analyzer

sambetts-cisco:
Add support for Trivy as an SBOM analyzer

This the first part of an effort to introduce Trivy support to
kubeclarity.

This commit adds Trivy as an SBOM analyzer which can be used in
conjunction with the other SBOM analyzers or by itself to scan the
contents of container images or directories and produce a merged result.

Along with Trivy we introduce a new input type "rootfs" which allows
kubeclarity to scan operating system root file systems like unpacked
containers or virtual machine volumes.


shared/pkg/analyzer/trivy/trivy.go
```diff
@@ -0,0 +1,125 @@
...
+			},
+			ScanOptions: trivyFlag.ScanOptions{
+				Target:         userInput,
+				SecurityChecks: nil, // Disable all security checks for SBOM only scan
+			},
+			ReportOptions: trivyFlag.ReportOptions{
+				Format:       "cyclonedx", // Cyconedx format for SBOM so that we don't need to convert
+				ReportFormat: "all",       // Full report not just summary
+				Output:       &output,     // Save the output to our local buffer instead of Stdout
+				ListAllPkgs:  true,        // List all packages for full SBOM
```
akpsgit, line 76: Needed to look what this means for trivy. maybe better to state "List all packages regardless of vulnerabilities"

shared/pkg/analyzer/trivy/trivy.go
```diff
@@ -0,0 +1,125 @@
...
+			return
+		}
+
+		err := artifact.Run(context.TODO(), trivyOptions, trivySourceType)
+		if err != nil {
+			a.setError(res, fmt.Errorf("failed to generate SBOM: %w", err))
+			return
+		}
+
+		frm := formatter.New(formatter.CycloneDXJSONFormat, output.Bytes())
```
akpsgit, line 102: shouldn't this be CycloneDX XML? i see that the merge sbom logic doing formatting, but maybe better to have a uniform return format?

sambetts-cisco:
Moved back to draft after identifying issues with standardising the cyclonedx from trivy, hashes etc are missing from the component.

shared/pkg/analyzer/trivy/trivy.go
sambetts-cisco: updated the comment

akpsgit (dismissed):

pbalogh-sa (dismissed):

pbalogh-sa (approved):

PR #254, opened 2022-11-11
Use input SBOM format directly as the input for grype

sambetts-cisco:
Previously we always converted to syft-json before giving it to grpye
and because of a bug in the syft conversion logic we were using it
resulted in a broken input for certain packages.

This commit updates the scanner logic so that which ever SBOM format we
receive from the user we pass it directly to grpye for scanning. This
does mean that if we are given a broken syft-json input then it will
still error, and we have a ticket open in the syft project to resolve
this.


shared/pkg/scanner/grype/local_grype.go
```diff
@@ -74,18 +74,11 @@ func (s *LocalScanner) run(sourceType utils.SourceType, userInput string) {
 	var hash string
 	origInput := userInput
 	if sourceType == utils.SBOM {
-		syftJSONFilePath, cleanup, err := ConvertCycloneDXFileToSyftJSONFile(userInput, s.logger)
```
FrimIdan, line 77: `ConvertCycloneDXFileToSyftJSONFile` is used also in `remote_grype` - we need to remove it also in that case and fix the code in the grype server to be able to handle any sbom formt?

FrimIdan:
@sambetts-cisco need to separate the PR from the trivy PR

shared/pkg/scanner/grype/common.go
```diff
@@ -233,7 +233,12 @@ func getOriginalInputAndHashFromSBOM(inputSBOMFile string) (string, string, erro
 	if err != nil {
 		return "", "", converter.ErrFailedToGetCycloneDXSBOM
 	}
-	hash := cdx_helper.GetComponentHash(cdxBOM.Metadata.Component)
+	var hash string
+	var name string
+	if cdxBOM.Metadata != nil && cdxBOM.Metadata.Component != nil {
+		name = cdxBOM.Metadata.Component.Name
+		hash = cdx_helper.GetComponentHash(cdxBOM.Metadata.Component)
```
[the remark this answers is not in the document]
sambetts-cisco: I've dropped this from the PR it was a result of using a rootfs scan result. I'll need to fix this in the Trivy PR.

shared/pkg/converter/cyclonedx_to_syft.go
```diff
@@ -74,23 +77,37 @@ func saveSyftSBOMToFile(syftBOM syft_sbom.SBOM, outputSBOMFile string) error {
...
-	// use the formatter
-	if err = input.Decode(inputFormat); err != nil {
-		return nil, fmt.Errorf("unable to decode input SBOM %s: %v", inputSBOMFile, err)
+
+	// Ensure input is converted to cyclonedx regardless of the
+	// input SBOM type.
+	r := bytes.NewReader(inputSBOMBytes)
+	sbom, format, err := syft.Decode(r)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read SBOM file %s: %w", inputSBOMFile, err)
```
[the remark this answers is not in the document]
sambetts-cisco: updated

shared/pkg/scanner/grype/local_grype.go
sambetts-cisco: removed the conversion from remote_grype too, thanks!

sambetts-cisco:
> And a general question, now that we know what's wrong with syft json after convert, can't we just set the value of the missing metadata until there will be a proper fix in syft?

I think its better to support the formats that the scanners support themselves directly without needing a conversion, it saves the system time and resources by not decoding/recoding the sbom twice, hopefully speeding things up, and also it removes the responsibility from our side to ensure all the metadata from one format is passed through to the other format.

shared/pkg/converter/cyclonedx_to_syft.go
```diff
@@ -74,23 +77,43 @@ func saveSyftSBOMToFile(syftBOM syft_sbom.SBOM, outputSBOMFile string) error {
...
+	// understand. Other format's metadata is best effort so we'll use
+	// syfts intermediary struct and then use ToFormatModel to switch it to
+	// cdx.BOM.
+	var bom *cdx.BOM
+	cdxFormat := cdx.BOMFileFormatXML
+	switch format {
+	case syft.FormatByName("cyclonedxjson"):
+		cdxFormat = cdx.BOMFileFormatJSON
+		fallthrough
+	case syft.FormatByName("cyclonedx"):
```
fishkerez, line 105: In the case of cyclonedxjson , you will decode it with the syft decoder. By the comment above you want to convert it using cdx.NewBOMDecoder. Maybe you did mean to put case syft.FormatByName("cyclonedx"), case syft.FormatByName("cyclonedxjson"): ?

e2e/cli_scan_test.go
```diff
@@ -80,6 +86,23 @@ func TestCLIScan(t *testing.T) {
 			scanImage(t, TestImageName, appID)
 			validateScanImage(t)
 
+			return ctx
+		}).
+		Assess("cli scan flow - image with known bad metadata in cyclonedx", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
```
fishkerez, line 91: The problem here is that the database already has results from previous test - "cli scan flow - analyze and scan"
You need to either clean up the database, or do a more precise validation (like component id or somthing)

shared/pkg/converter/cyclonedx_to_syft.go
sambetts-cisco: the "fallthough" from cyclonedxjson falls through to the cyclonedx case after changing the cdxFormat variable, it doesn't fallthrough to the syft FormatToModel case in the switch statement

e2e/cli_scan_test.go
sambetts-cisco: I'm using a different application name and ID I thought that would be enough to separate the validation, I need to re-read the validate functions

shared/pkg/converter/cyclonedx_to_syft.go
fishkerez: But it will never enter to the cyclonedx case if it entered to the cyclonedxjson case, no?

sambetts-cisco: "fallthrough" will results in the code flow moving into the next case regardless of the condition

fishkerez: Oh I didn't know that.. thought it is just checking it  👍

FrimIdan (approved):

e2e/cli_scan_test.go
sambetts-cisco: We're now filtering all the validate steps using the appID unique to this test so I think we should be good.

FrimIdan (approved):

PR #262, opened 2022-11-25
Improve analyzer SBOM merging

sambetts-cisco:
This PR includes several commits to improve the analyzer SBOM merging logic, removing the need for extra encode/decode operations, and also handling some of the more complex fields in the components so that information isn't lost during the merge.


pbalogh-sa (approved):
LGTM

PR #263, opened 2022-11-27
Job manager changes

FrimIdan:
Introducing new job factory logic needed from job manager to be more generic and not coupled with kubeclarity configs. 
Also, struct labels was added to all configs.

### DIFF
--- a/cli/pkg/config/config.go
+++ b/cli/pkg/config/config.go
@@ -45,9 +45,9 @@ func LoadConfig() *Config {
 		Backend:       loadBackendConfig(),
 		TimeOut:       viper.GetDuration(TimeOut),
 		SharedConfig: &shared.Config{
-			Registry:       loadRegistryConfig(),
-			Analyzer:       shared.LoadAnalyzerConfig(),
-			Scanner:        shared.LoadScannerConfig(),
+			Registry:       *loadRegistryConfig(),
+			Analyzer:       *shared.LoadAnalyzerConfig(),
+			Scanner:        *shared.LoadScannerConfig(),
 			LocalImageScan: viper.GetBool(shared.LocalImageScan),
 		},
 	}
--- a/cli/pkg/config/config.go
+++ b/cli/pkg/config/config.go
@@ -33,7 +33,7 @@ type Config struct {
 	LogLevel      string
 	EnableJSONLog bool
 	Backend       *Backend
-	SharedConfig  shared.Config
+	SharedConfig  *shared.Config
 	TimeOut       time.Duration
 }
 
@@ -44,10 +44,10 @@ func LoadConfig() *Config {
 		EnableJSONLog: viper.GetBool(EnableJSONLog),
 		Backend:       loadBackendConfig(),
 		TimeOut:       viper.GetDuration(TimeOut),
-		SharedConfig: shared.Config{
-			Registry:       *loadRegistryConfig(),
-			Analyzer:       *shared.LoadAnalyzerConfig(),
-			Scanner:        *shared.LoadScannerConfig(),
+		SharedConfig: &shared.Config{
+			Registry:       loadRegistryConfig(),
+			Analyzer:       shared.LoadAnalyzerConfig(),
+			Scanner:        shared.LoadScannerConfig(),
 			LocalImageScan: viper.GetBool(shared.LocalImageScan),
 		},
 	}
--- a/cli/pkg/config/config.go
+++ b/cli/pkg/config/config.go
@@ -33,7 +33,7 @@ type Config struct {
 	LogLevel      string
 	EnableJSONLog bool
 	Backend       *Backend
-	SharedConfig  *shared.Config
+	SharedConfig  shared.Config
 	TimeOut       time.Duration
 }
 
@@ -44,7 +44,7 @@ func LoadConfig() *Config {
 		EnableJSONLog: viper.GetBool(EnableJSONLog),
 		Backend:       loadBackendConfig(),
 		TimeOut:       viper.GetDuration(TimeOut),
-		SharedConfig: &shared.Config{
+		SharedConfig: shared.Config{
 			Registry:       *loadRegistryConfig(),
 			Analyzer:       *shared.LoadAnalyzerConfig(),
 			Scanner:        *shared.LoadScannerConfig(),
--- a/cli/pkg/utils/sbom.go
+++ b/cli/pkg/utils/sbom.go
@@ -1,3 +1,18 @@
+// Copyright © 2022 Cisco Systems, Inc. and its affiliates.
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package utils
 
 import (
--- a/cli/pkg/utils/sbom.go
+++ b/cli/pkg/utils/sbom.go
@@ -1,0 +1,31 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/openclarity/kubeclarity/shared/pkg/converter"
+	"github.com/openclarity/kubeclarity/shared/pkg/formatter"
+)
+
+func ConvertInputSBOMIfNeeded(inputSBOMFile, outputFormat string) ([]byte, error) {
+	inputSBOM, err := os.ReadFile(inputSBOMFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read SBOM file %s: %v", inputSBOMFile, err)
+	}
+	inputSBOMFormat := converter.DetermineCycloneDXFormat(inputSBOM)
+	if inputSBOMFormat == outputFormat {
+		return inputSBOM, nil
+	}
+
+	// Create cycloneDX formatter to convert input SBOM to the defined output format.
+	cdxFormatter := formatter.New(inputSBOMFormat, inputSBOM)
+	if err = cdxFormatter.Decode(inputSBOMFormat); err != nil {
+		return nil, fmt.Errorf("failed to decode input SBOM %s: %v", inputSBOMFile, err)
+	}
+	if err := cdxFormatter.Encode(outputFormat); err != nil {
+		return nil, fmt.Errorf("failed to encode input SBOM: %v", err)
+	}
+
+	return cdxFormatter.GetSBOMBytes(), nil
+}
--- a/shared/pkg/analyzer/job/job.go
+++ b/shared/pkg/analyzer/job/job.go
@@ -25,14 +25,15 @@ import (
 	"github.com/openclarity/kubeclarity/shared/pkg/job_manager"
 )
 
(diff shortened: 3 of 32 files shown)


shared/pkg/analyzer/job/job.go
```diff
@@ -18,24 +18,20 @@ package job
...
-	case cdx_gomod.AnalyzerName:
-		return cdx_gomod.New(config, logger, resultChan)
-	case trivy.AnalyzerName:
-		return trivy.New(config, logger, resultChan)
-	default:
-		logger.Fatalf("Unknown analyzer: %v", analyzerName)
+func CreateAnalyzerJob(scannerName string, conf job_manager.IsConfig, logger *logrus.Entry, resultChan chan job_manager.Result) job_manager.Job {
+	createJobFunc, ok := job_factory.GetCreateJobFuncs()[scannerName]
+	if !ok {
+		logrus.Fatalf("Unregister analyzer: %v", scannerName)
```
akpsgit, line 34: Unregistered

shared/pkg/config/analyzer.go
```diff
@@ -23,10 +23,10 @@ import (
...
 // TODO: maybe we need to extend the unified config.
 type Analyzer struct {
-	OutputFormat string
-	AnalyzerList []string
-	Scope        string
-	TrivyConfig  AnalyzerTrivyConfig
+	OutputFormat string              `yaml:"output_format" mapstructure:"output_format"`
+	AnalyzerList []string            `yaml:"analyzer_list" mapstructure:"analyzer_list"`
+	Scope        string              `yaml:"scope" mapstructure:"scope"`
+	TrivyConfig  AnalyzerTrivyConfig `yaml:"trivy_config" mapstructure:"trivy_config"`
```
akpsgit, line 29: Shouldn't this be a dynamic config and not part of the infrastructure?

shared/pkg/scanner/job/job.go
```diff
@@ -18,21 +18,20 @@ package job
...
-	case grype.ScannerName:
-		return grype.New(config, logger, resultChan)
-	case dependency_track.ScannerName:
-		return dependency_track.New(config, logger, resultChan)
-	default:
-		logrus.Fatalf("Unknown scanner: %v", scannerName)
+func CreateJob(scannerName string, conf job_manager.IsConfig, logger *logrus.Entry, resultChan chan job_manager.Result) job_manager.Job {
+	createJobFunc, ok := job_factory.GetCreateJobFuncs()[scannerName]
+	if !ok {
+		logrus.Fatalf("Unregister scanner: %v", scannerName)
```
akpsgit, line 34: Unregistered

cli/pkg/config/config.go
```diff
@@ -44,10 +44,10 @@ func LoadConfig() *Config {
 		EnableJSONLog: viper.GetBool(EnableJSONLog),
 		Backend:       loadBackendConfig(),
 		TimeOut:       viper.GetDuration(TimeOut),
-		SharedConfig: &shared.Config{
-			Registry:       loadRegistryConfig(),
-			Analyzer:       shared.LoadAnalyzerConfig(),
-			Scanner:        shared.LoadScannerConfig(),
+		SharedConfig: shared.Config{
```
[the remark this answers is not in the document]
FrimIdan: I've undo all pointer changes, they were not needed.
see https://github.com/openclarity/kubeclarity/pull/263/commits/0c2aa0b430fac70cf5707c783bb003f48daf1347

cli/pkg/utils/sbom.go
```diff
@@ -0,0 +1,46 @@
...
+	cdxFormatter := formatter.New(inputSBOMFormat, inputSBOM)
+	if err = cdxFormatter.Decode(inputSBOMFormat); err != nil {
+		return nil, fmt.Errorf("failed to decode input SBOM %s: %v", inputSBOMFile, err)
+	}
+	if err := cdxFormatter.Encode(outputFormat); err != nil {
+		return nil, fmt.Errorf("failed to encode input SBOM: %v", err)
+	}
+
+	return cdxFormatter.GetSBOMBytes(), nil
+}
```
sambetts-cisco, line 46: This function is being completely removed in my PR: https://github.com/openclarity/kubeclarity/pull/262 maybe we should wait to add/move this until all those things land?

shared/pkg/analyzer/trivy/trivy.go
```diff
@@ -42,10 +43,12 @@ type Analyzer struct {
 	localImage bool
 }
 
-func New(conf *config.Config,
-	logger *log.Entry,
-	resultChan chan job_manager.Result,
-) job_manager.Job {
+func init() {
+	job_factory.RegisterCreateJobFunc(AnalyzerName, New)
```
sambetts-cisco, line 47: Right now we have a global job factory, but I think we need a per-family job factory, the trivy analyzer should only be registering as a job within the analyzer space.

The current implementation here will run into issues when there are analyzers/scanners with the same name for example analyzer/trivy and scanner/trivy. Also there is nothing to prevent me accidentally enabling a scanner as part of an analyzer run.

I think it would make much more sense if analyzer/job looked like:

```
import (
_ "github.com/openclarity/kubeclarity/shared/pkg/analyzer/cdx_gomod"
_ "github.com/openclarity/kubeclarity/shared/pkg/analyzer/syft"
_ "github.com/openclarity/kubeclarity/shared/pkg/analyzer/trivy"
)

var JobFactory = &job_factory.Factory{}
```

trivy would look like:

```
import "github.com/openclarity/kubeclarity/shared/pkg/analyzer"

func init() {
   analyzer.JobFactory.Register(ScannerName, New) 
}
```

and then job_manager looked like:

```
func New(jobNames []string, config IsConfig, logger *logrus.Entry, jobFactory *job_factory.Factory) *Manager {
}

func (m *Manager) Run(...) {
        ...
        
     	jobs := make([]Job, len(m.jobNames))
	for i, name := range m.jobNames {
		nameToResultChan[name] = make(chan Result, 10) // nolint:gomnd
		jobs[i], err = m.jobFactory.CreateJob(name, m.config, m.logger, nameToResultChan[name])
                if err != nil {
                    log.Fatalf("unable to build list of jobs: %w": err)
                }
	}
	
	...
}
```

and cli/analyze and runtime_k8s_scanner/analyze look like:

```
manager := job_manager.New(..., analyzer.JobFactory)
```

shared/pkg/job_manager/config.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package job_manager // nolint:revive,stylecheck
+
+type IsConfig interface {
+	IsConfig()
+}
```
sambetts-cisco, line 20: I don't really get the reasoning behind this interface, I think we should either statically type the config with a top level struct which all the of the jobs accept and then pull out the stuff they care about, or use a *Viper instead to remove the need for a top level struct.

shared/pkg/job_factory/factory.go
```diff
@@ -0,0 +1,45 @@
...
+	})
+	if _, ok := createJobFuncs[name]; ok {
+		logrus.Fatalf("%q already registered", name)
+	}
+	createJobFuncs[name] = createJobFunc
+}
+
+func GetCreateJobFuncs() map[string]CreateJobFunc {
+	return createJobFuncs
+}
```
sambetts-cisco, line 45: See my other comment for more details but I think this should look something like:
```
type Factory struct {
    jobFuncs map[string]CreateJobFunc
}

func (f *Factory) Register(name string, createJobFunc CreateJobFunc) {
    if f.jobFuncs == nil {
        f.jobFuncs = map[string]CreateJobFunc{}
    }
    if _, ok := f.jobFuncs[name]; ok {
	logrus.Fatalf("%q already registered", name)
    } 
    f.jobFuncs[name] = createJobFunc
}

func (f *Factory) CreateJob(name string, conf job_manager.IsConfig, logger *logrus.Entry, resultChan chan job_manager.Result) (job_manager.Job, error) {
    createFunc, ok := f.jobFuncs[name]
    if !ok {
       return nil, fmt.Errorf("%v not a registered job", name)
    }
    return createFunc(conf, logger, resultChan)
} 
```

sambetts-cisco (changes requested):

shared/pkg/analyzer/trivy/trivy.go
FrimIdan: @sambetts-cisco I think that i've understand, but the problem that I see is that we have a cycle import here.
From your example, if trivy needs to import `analyzer` who will import `analyzer/job` for the all empty imports of the different analyzers will kick in?
unless I will do an empty import in each main that will use the job manager I'm not sure how I can solve it - I will give it a try.

FrimIdan: Also I have a cycle import when `job_manager` --> `job_factory` -->  `job_manager`

We might want to open a task to make a better refactor once we finalize the exact structure. meanwhile we can just call the scanner trivy and the analyzer trivy with a different name. WDYT?

sambetts-cisco: So the job_manager --> job_factory --> job_manager import cycle can be fixed by moving the Job definition to job factory, which is probably were it belongs anyway, or just moving the definition of Factory into job_manager as they are tightly coupled.

The analyzer import loop is by the directionality of our job factory registration, and we can't really get around that without having to add two imports in cli one for the job factory location and one too trigger all the registrations.

I think the way we're registering the jobs isn't recommended from a golang perspective, instead Registering to the job factory in the analyzer/job package would be better like:

```
import (
 "github.com/openclarity/kubeclarity/shared/pkg/analyzer/cdx_gomod" "github.com/openclarity/kubeclarity/shared/pkg/analyzer/syft"
 "github.com/openclarity/kubeclarity/shared/pkg/analyzer/trivy"
)

var JobFactory = &job_factory.Factory{}

func init() {
    JobFactory.Register(trivy.ScannerName, trivy.New)
    JobFactory.Register(syft.ScannerName, syft.New)
    JobFactory.Register(cdx_gomod.ScannerName, cdx_gomod.New)
}
```

alternatively we can move the factory to the analyzer package:

analyzer/analyzer.go
```
package analyzer

var JobFactory = &job_factory.Factory{}
...
```

analyzer/jobs/jobs.go
```
package jobs

import (
 _ "github.com/openclarity/kubeclarity/shared/pkg/analyzer/cdx_gomod"
 _ "github.com/openclarity/kubeclarity/shared/pkg/analyzer/syft"
 _ "github.com/openclarity/kubeclarity/shared/pkg/analyzer/trivy"
)
```

analyzer/trivy/common.go
```
import "analyzer"

func init() {
   analyzer.JobFactory.Register(ScannerName, New) 
}
```

but then cli/cmd/analyze needs to do both:

```
import _ "analyzer/jobs" // triggers all the job registrations to the job factory
import "analyzer" // to get the job factory to pass to the job_manager
```

personal opinion is that the first option is better.

sambetts-cisco (commented):
just need to decide how we want to refactor the job factory, we can do it now, or we can do it in a follow up, I'm happy to take on the refactor if there is time

cli/pkg/utils/sbom.go
FrimIdan: whoever push first? I really don't care

shared/pkg/analyzer/trivy/trivy.go
FrimIdan: thanks @sambetts-cisco I think that https://github.com/openclarity/kubeclarity/pull/263/commits/846d884b4eb7fc121e1b97afa5b8e9f768a7844e address it

sambetts-cisco (approved):

PR #270, opened 2022-12-05
Add trivy support to kubeclarity helm chart

sambetts-cisco:
This commit adds support for the trivy analyzer and scanner to the kubeclarity helm chart including trivy remote mode to ensure that we centralize the trivy DB and cache management.


FrimIdan (approved):
lgtm

charts/kubeclarity/templates/scanner-template-configmap.yaml
```diff
@@ -71,6 +71,20 @@ data:
               value: {{ include "kubeclarity.grype-server.name" . }}.{{ .Release.Namespace }}:{{ index .Values "kubeclarity-grype-server" "servicePort" }}
             - name: SCANNER_REMOTE_GRYPE_SERVER_TIMEOUT
               value: {{ index .Values "kubeclarity-runtime-scan" "vulnerability-scanner" "scanner" "grype" "remote-grype" "timeout" }}
+{{- end }}
+            - name: ANALYZER_TRIVY_CACHE_DIRECTORY
```
FrimIdan, line 75: shouldn't `ANALYZER_TRIVY_CACHE_DIRECTORY` be configured only if trivy analyzer is enabled?

charts/kubeclarity/templates/scanner-template-configmap.yaml
```diff
@@ -71,6 +71,20 @@ data:
...
+              value: /tmp
+{{- if index .Values "kubeclarity-runtime-scan" "vulnerability-scanner" "analyzer" "trivy" "enabled" }}
+            - name: ANALYZER_TRIVY_TIMEOUT
+              value: {{ index .Values "kubeclarity-runtime-scan" "vulnerability-scanner" "analyzer" "trivy" "timeout" }}
+{{- end }}
+{{- if index .Values "kubeclarity-runtime-scan" "vulnerability-scanner" "scanner" "trivy" "enabled" }}
+            - name: SCANNER_TRIVY_TIMEOUT
```
FrimIdan, line 82: what about `SCANNER_TRIVY_CACHE_DIRECTORY`?

charts/kubeclarity/templates/trivy_server/deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+            periodSeconds: 30
+            failureThreshold: 5
+            timeoutSeconds: 10
+          livenessProbe:
+            tcpSocket:
+              port: {{ index .Values "kubeclarity-trivy-server" "servicePort" }}
+            initialDelaySeconds: 10
+            periodSeconds: 30
+            failureThreshold: 5
+            timeoutSeconds: 10
```
FrimIdan, line 61: based on https://github.com/aquasecurity/trivy/issues/688 I think we can use the `/healthz` (https://github.com/aquasecurity/trivy/blob/main/pkg/rpc/server/listen.go#L96-L100) and not use tcpSocket.
Maybe we should even use the existing [chart](https://github.com/aquasecurity/trivy/tree/main/helm/trivy) as a dependency. 

Not sure if we need more from the trivy chart here

sambetts-cisco: weird they don't document their chart anywhere on their actual docs page only in the repo so I missed it, I'll see if I can  add it as a dependency it'll give us a lot more configuration options

charts/kubeclarity/templates/scanner-template-configmap.yaml
```diff
@@ -92,10 +109,10 @@ data:
 {{- end}}
 {{- if index .Values "kubeclarity-runtime-scan" "httpsProxy" }}
             - name: NO_PROXY
-              value: {{ include "kubeclarity.name" . }}.{{ .Release.Namespace }}:{{ index .Values "kubeclarity-runtime-scan" "resultServicePort" }},{{ include "kubeclarity.grype-server.name" . }}.{{ .Release.Namespace }}:{{ index .Values "kubeclarity-grype-server" "servicePort" }}
+              value: {{ include "kubeclarity.name" . }}.{{ .Release.Namespace }}:{{ index .Values "kubeclarity-runtime-scan" "resultServicePort" }},{{ include "kubeclarity.grype-server.name" . }}.{{ .Release.Namespace }}:{{ index .Values "kubeclarity-grype-server" "servicePort" }},{{ include "kubeclarity.trivy-server.name" . }}.{{ .Release.Namespace }}:{{ index .Values "kubeclarity-trivy-server" "servicePort" }}
```
FrimIdan, line 112: `servicePort`? 
is it not `"service" "port"`?

sambetts-cisco: great catch, thanks!

shared/pkg/config/trivy.go
```diff
@@ -23,10 +23,14 @@ import (
...
-const AnalyzerTrivyTimeoutSeconds = "ANALYZER_TRIVY_TIMEOUT_SECONDS"
+const (
+	AnalyzerTrivyTimeoutSeconds = "ANALYZER_TRIVY_TIMEOUT_SECONDS"
+	AnalyzerTrivyCacheDir       = "ANALYZER_TRIVY_CACHE_DIRECTORY"
+)
 
 type AnalyzerTrivyConfig struct {
-	Timeout int `yaml:"timeout" mapstructure:"timeout"`
+	Timeout  int    `yaml:"timeout" mapstructure:"timeout"`
+	CacheDir string `yaml:"cache-dir" mapstructure:"cache-dir"`
```
FrimIdan, line 33: can it be `cache_dir` and not `cache-dir`?
I think we are using underscore in all other places

FrimIdan (approved):

PR #276, opened 2022-12-14
helm: adding the possibility to create an ingress

Mesh33:
Hello everyone. 

I know there is no authentication yet on Kubeclarity, but I needed an ingress to be able to reach my Kubeclarity instance through an hostname, only reachable when connected to a VPN.

I tried to make retrocompatible changes and it shouldn't impact current helm deployments.

I tried this on my side (GKE) and can now:
- Access the UI with an hostname
- Access the backend in my CI/CD pipelines through the hostname

On my environment, a VPN is needed to reach the Kubeclarity instance.

Please let me know modifications I may have to make. 

Thanks everyone for your comments.

### DIFF
--- a/charts/kubeclarity/templates/ingress.yaml
+++ b/charts/kubeclarity/templates/ingress.yaml
@@ -1,0 +1,56 @@
+{{- if .Values.kubeclarity.ingress.enabled -}}
+{{- $kubeVersion := .Capabilities.KubeVersion.Version }}
+{{- $pathType := .Values.kubeclarity.ingress.pathType -}}
+{{- $serviceName := include "kubeclarity.name" . -}}
+{{- $servicePort := .Values.kubeclarity.service.port -}}
+{{ if semverCompare ">= 1.19.0-0" $kubeVersion }}
+apiVersion: networking.k8s.io/v1
+{{ else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
+apiVersion: networking.k8s.io/v1beta1
+{{ else }}
+apiVersion: extensions/v1beta1
+{{ end }}
+kind: Ingress
+metadata:
+  name: {{ include "kubeclarity.name" . }}
+  namespace: '{{ .Release.Namespace }}'
+  labels:
+    {{ include "kubeclarity.labels" . }}
+  {{- template "kubeclarity.ingress.annotations" . }}
+spec:
+{{- if .Values.kubeclarity.ingress.tls }}
+  tls:
+  {{- range .Values.kubeclarity.ingress.tls }}
+    - hosts:
+      {{- range .hosts }}
+        - {{ . | quote }}
+      {{- end }}
+      secretName: {{ .secretName }}
+  {{- end }}
+{{- end }}
+{{- if .Values.kubeclarity.ingress.ingressClassName }}
+  ingressClassName: {{ .Values.kubeclarity.ingress.ingressClassName }}
+{{- end }}
+  rules:
+  {{- range .Values.kubeclarity.ingress.hosts }}
+    - host: {{ .host | quote }}
+      http:
+        paths:
+        {{- range (.paths | default (list "/")) }}
+          - path: {{ . -}}
+            {{- if semverCompare ">= 1.19.0-0" $kubeVersion }}
+            pathType: {{ $pathType }}
+            {{- end }}
+            backend:
+              {{- if semverCompare ">= 1.19.0-0" $kubeVersion }}
+              service:
+                name: {{ $serviceName }}
+                port:
+                  number: {{ $servicePort }}
+              {{ else }}
+              serviceName: {{ $serviceName }}
+              servicePort: {{ $servicePort }}
+              {{ end }}
+        {{- end }}
+  {{- end }}
+{{- end }}
--- a/charts/kubeclarity/templates/ingress.yaml
+++ b/charts/kubeclarity/templates/ingress.yaml
@@ -1,6 +1,5 @@
 {{- if .Values.kubeclarity.ingress.enabled -}}
 {{- $kubeVersion := .Capabilities.KubeVersion.Version }}
-{{- $pathType := .Values.kubeclarity.ingress.pathType -}}
 {{- $serviceName := include "kubeclarity.name" . -}}
 {{- $servicePort := 8080 -}}
 {{ if semverCompare ">= 1.19.0-0" $kubeVersion }}
@@ -36,10 +35,10 @@ spec:
     - host: {{ .host | quote }}
       http:
         paths:
-        {{- range (.paths | default (list "/")) }}
-          - path: {{ . -}}
+        {{- range (.paths | default (list (dict "pathType" "Prefix" "path" "/"))) }}
+          - path: {{ .path -}}
             {{- if semverCompare ">= 1.19.0-0" $kubeVersion }}
-            pathType: {{ $pathType }}
+            pathType: {{ .pathType }}
             {{- end }}
             backend:
               {{- if semverCompare ">= 1.19.0-0" $kubeVersion }}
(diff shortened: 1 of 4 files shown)


charts/kubeclarity/templates/ingress.yaml
```diff
@@ -0,0 +1,60 @@
...
+{{ end }}
+kind: Ingress
+metadata:
+  name: {{ include "kubeclarity.name" . }}
+  namespace: '{{ .Release.Namespace }}'
+  labels:
+    {{ include "kubeclarity.labels" . }}
+  {{- template "kubeclarity.ingress.annotations" . }}
+spec:
+{{- if .Values.kubeclarity.ingress.enabled }}
```
FrimIdan, line 22: did you mean `{{- if .Values.kubeclarity.ingress.tls }}`

charts/kubeclarity/values.yaml
```diff
@@ -42,6 +42,35 @@ kubeclarity:
...
+
+    # As of Kubernetes 1.19, all Ingress Paths must have a pathType configured. The default value below should be sufficient in most cases.
+    # See: https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types for other possible values.
+    pathType: Prefix
+
+    hosts:
+      - host: chart-example.local
+        paths: []
+    ## Extra paths to prepend to the host configuration. This is useful when working with annotation based services.
+    extraPaths: []
```
sambetts-cisco, line 62: Looking at the template I think this is supposed to be indented along with paths, otherwise we're adding all extraPaths to every host. If that is the intention then I think we need to document here with a comment that extra paths will be prepended to *every* configured host's paths.

sambetts-cisco (changes requested):
Hi @Mesh33, Thanks for the PR! Optionally supporting ingress in the helm chart seems like a good idea, we need to make sure its very obvious that if you expose kubeclarity through ingress that there is no authentication and therefore open for anyone to access.

If you are trying to access kubeclarity from another kubernetes cluster, which parts of the system do you need exposed? Is it just the UI/API?

Could you fill out the commit message a bit more, explaining what the change is and why its required. Also right now we're not using conventional commit messages, so a more human readable commit headline would be great like: Add support for ingress in helm chart.

charts/kubeclarity/templates/ingress.yaml
Mesh33: That's right, corrected.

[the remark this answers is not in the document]
Mesh33: That's right, corrected.

charts/kubeclarity/values.yaml
```diff
@@ -42,6 +42,35 @@ kubeclarity:
...
+    # - path: /*
+    #   backend:
+    #     service:
+    #       name: ssl-redirect
+    #       port:
+    #         number: use-annotation
+    tls: []
+    #  - secretName: chart-example-tls
+    #    hosts:
+    #      - chart-example.local
```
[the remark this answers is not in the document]
Mesh33: I modified the ingress template to be less generic.
Also, I added the service port and targetPort in values.yaml, so it can be imported in both service.yaml and ingress.yaml templates.

charts/kubeclarity/values.yaml
Mesh33: I removed extraPaths as it was too generic.

Mesh33:
> Hi @Mesh33, Thanks for the PR! Optionally supporting ingress in the helm chart seems like a good idea, we need to make sure its very obvious that if you expose kubeclarity through ingress that there is no authentication and therefore open for anyone to access.
> 
> If you are trying to access kubeclarity from another kubernetes cluster, which parts of the system do you need exposed? Is it just the UI/API?
> 
> Could you fill out the commit message a bit more, explaining what the change is and why its required. Also right now we're not using conventional commit messages, so a more human readable commit headline would be great like: Add support for ingress in helm chart.

Hello @sambetts-cisco 
Thanks a lot for your time and your comments. I'm sorry for the mistakes I did as it is one of my first helm PRs ever.

I tried to take into account your comments, and also changed all my commits to be more descriptives. 

Have a good day.

charts/kubeclarity/values.yaml
```diff
@@ -41,6 +41,36 @@ kubeclarity:
 
   service:
     type: ClusterIP
+    annotations: {}
+    # Port on which Kubeclarity server is listening
+    port: 8080
+    # Target port to which the service should be mapped to
+    targetPort: 8080
```
sambetts-cisco, line 48: The target port is defined by the kubeclarity deployment (i.e. its the port that the kubeclarity pod is exposing) and this isn't something that is currently configurable so setting this to anything other than 8080 will break the service.

If we want to be able to change this like we do for the runtime scan port, then I think we need a configuration outside of "service" for listenPort and then we need to template it into the deployment.yaml.

IMO Its not something that I think needs to be configurable, the fact that kubeclarity is being exposed on 8080 behind the service port (which is user facing and can be/should configurable) is pretty much an entirely internal detail.

charts/kubeclarity/values.yaml
```diff
@@ -41,6 +41,36 @@ kubeclarity:
...
+
+    # As of Kubernetes 1.19, all Ingress Paths must have a pathType configured. The default value below should be sufficient in most cases.
+    # See: https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types for other possible values.
+    pathType: Prefix
+
+    hosts:
+        # hostname you want to use
+      - host: chart-example.local
+        # path will default to /
+        paths: []
```
sambetts-cisco, line 69: given we only have a single backend endpoint/service, do we have a use case for exposing the same endpoint on different paths, vs just having:
```
hosts:
  - host: chart-example.local
     path: /foo
```
I was also thinking that it might be worth exposing the pathType per path too like:
```
hosts:
  - host: chart-example.local
    # defaults to /
    path: /foo
    #defaults to Prefix
    pathType: Prefix 
```
or if we do need a list:
```
hosts:
  - host: chart-example.local
     paths: 
        - path: /foo # defaults to /
          pathType: Prefix # defaults to Prefix
```

sambetts-cisco (changes requested):
Thanks for the update, this is looking much better, I've got a couple more comments but I think this looks pretty close.

charts/kubeclarity/values.yaml
```diff
@@ -42,6 +42,35 @@ kubeclarity:
...
     type: ClusterIP
 
+  ingress:
+    enabled: false
+    labels: {}
+    annotations: {}
+
+    # Optionally use ingressClassName instead of deprecated annotation.
+    # See: https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation
+    ingressClassName: ""
```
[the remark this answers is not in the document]
Mesh33: ingressClassName is set to empty string by default

charts/kubeclarity/values.yaml
Mesh33: Done, paths kept as a list with pathType configurable

charts/kubeclarity/values.yaml
```diff
@@ -41,6 +41,31 @@ kubeclarity:
...
+
+    hosts:
+        # hostname you want to use
+      - host: chart-example.local
+        # path will default to /
```
sambetts-cisco, line 60: nit: a slight better comment about the default:
```
# paths will default to:
# - paths:
#    - pathType: Prefix
#      path: /
paths: []
```

sambetts-cisco (approved):
Tested myself, and everything appears to get setup correctly, but I'm only able to exercise the simplest cases. Templates all LGTM. Thanks!

charts/kubeclarity/values.yaml
Mesh33: Updated

sambetts-cisco (approved):
Thanks @Mesh33!

PR #277, opened 2022-12-14
helm: add the possibility to connect to an external PostgreSQL instance

Mesh33:
Hello everyone.
I wanted to use an external PostgreSQL database rather than the one used in dependencies, so I can decorrelate Kubeclarity storage from my Kubernetes cluster.

The goal of this PR is to use an external PostgreSQL database whereas currently, only using the one deployed in dependencies is possible. 

I tried to make retrocompatible changes.

### DIFF
--- a/charts/kubeclarity/Chart.yaml
+++ b/charts/kubeclarity/Chart.yaml
@@ -15,7 +15,7 @@ dependencies:
   - name: postgresql
     version: "11.6.12"
     repository: https://charts.bitnami.com/bitnami
-    condition: kubeclarity-postgresql.internal.enabled
+    condition: kubeclarity-postgresql.enabled
     alias: kubeclarity-postgresql
   - name: trivy
     version: "0.4.17"
--- a/charts/kubeclarity/Chart.yaml
+++ b/charts/kubeclarity/Chart.yaml
@@ -15,7 +15,7 @@ dependencies:
   - name: postgresql
     version: "11.6.12"
     repository: https://charts.bitnami.com/bitnami
-    condition: kubeclarity-postgresql.enabled
+    condition: kubeclarity-postgresql.internal.enabled
     alias: kubeclarity-postgresql
   - name: trivy
     version: "0.4.17"
--- a/charts/kubeclarity/templates/deployment.yaml
+++ b/charts/kubeclarity/templates/deployment.yaml
@@ -1,6 +1,6 @@
 {{- if index .Values "kubeclarity-postgresql" "internal" "enabled" }}
-{{- if index .Values "kubeclarity-postgresql-external" "enabled" }}
-{{- fail "kubeclarity-postgresql and kubeclarity-postgresql-external can not both be enabled!" }}
+{{- if index .Values "kubeclarity-postgresql" "external" "enabled" }}
+{{- fail "kubeclarity-postgresql.internal and kubeclarity-postgresql.external can not both be enabled!" }}
 {{- end }}
 {{- end }}
 {{- $dbHost := print (include "kubeclarity.name" .) "-postgresql" -}}
@@ -9,12 +9,12 @@
 {{- $dbName := index .Values "kubeclarity-postgresql" "internal" "auth" "database" -}}
 {{- $secretName := index .Values "kubeclarity-postgresql" "internal" "auth" "existingSecret" -}}
 {{- $secretKey := index .Values "kubeclarity-postgresql" "internal" "auth" "secretKey" -}}
-{{ if index .Values "kubeclarity-postgresql-external" "enabled" }}
-  {{- $dbHost = index .Values "kubeclarity-postgresql-external" "auth" "host" -}}
-  {{- $dbPort = index .Values "kubeclarity-postgresql-external" "auth" "port" | quote -}}
-  {{- $dbUser = index .Values "kubeclarity-postgresql-external" "auth" "user" -}}
-  {{- $dbName = index .Values "kubeclarity-postgresql-external" "auth" "database" -}}
-  {{- $secretName = index .Values "kubeclarity-postgresql-external" "auth" "existingSecret" -}}
+{{ if index .Values "kubeclarity-postgresql" "external" "enabled" }}
+  {{- $dbHost = index .Values "kubeclarity-postgresql" "external" "auth" "host" -}}
+  {{- $dbPort = index .Values "kubeclarity-postgresql" "external" "auth" "port" | quote -}}
+  {{- $dbUser = index .Values "kubeclarity-postgresql" "external" "auth" "user" -}}
+  {{- $dbName = index .Values "kubeclarity-postgresql" "external" "auth" "database" -}}
+  {{- $secretName = index .Values "kubeclarity-postgresql" "external" "auth" "existingSecret" -}}
   {{- $secretKey := index .Values "kubeclarity-postgresql" "internal" "auth" "secretKey" -}}
 {{ end }}
 apiVersion: apps/v1
@@ -136,9 +136,9 @@ spec:
               value: {{ $dbPort }}
             - name: DB_USER
               value: {{ $dbUser }}
-            {{- if index .Values "kubeclarity-postgresql-external" "auth" "password" }}
+            {{- if index .Values "kubeclarity-postgresql" "external" "auth" "password" }}
             - name: DB_PASS
-              value: {{ index .Values "kubeclarity-postgresql-external" "auth" "password" }}
+              value: {{ index .Values "kubeclarity-postgresql" "external" "auth" "password" }}
             {{- else }}
             - name: DB_PASS
               valueFrom:
--- a/charts/kubeclarity/templates/deployment.yaml
+++ b/charts/kubeclarity/templates/deployment.yaml
@@ -6,7 +6,7 @@
 {{- $secretKey := index .Values "kubeclarity-postgresql-secret" "secretKey" -}}
 {{- $secretName := index .Values "kubeclarity-postgresql" "auth" "existingSecret" -}}
 {{- $dbHost := print (include "kubeclarity.name" .) "-postgresql" -}}
-{{- $dbPort := index .Values "kubeclarity-postgresql" "auth" "port" | quote -}}
+{{- $dbPort := index .Values "kubeclarity-postgresql" "service" "ports" "postgresql" | quote -}}
 {{- $dbUser := index .Values "kubeclarity-postgresql" "auth" "user" -}}
 {{- $dbName := index .Values "kubeclarity-postgresql" "auth" "database" -}}
 {{ if index .Values "kubeclarity-postgresql-external" "enabled" }}
--- a/charts/kubeclarity/templates/deployment.yaml
+++ b/charts/kubeclarity/templates/deployment.yaml
@@ -40,7 +40,7 @@ spec:
       initContainers:
         - name: '{{ include "kubeclarity.name" . }}-wait-for-pg-db'
           image: {{ index .Values "kubeclarity-postgresql" "image" "registry" }}/{{ index .Values "kubeclarity-postgresql" "image" "repository" }}:{{ index .Values "kubeclarity-postgresql" "image" "tag" }}
-          command: ['sh', '-c', 'until pg_isready -h {{ $dbHost }} -p {{ $dbPort }} -U {{ $dbUser }} -d "dbname={{ $dbName }}";
+          command: ['sh', '-c', 'until pg_isready -h {{ $dbHost | quote }} -p {{ $dbPort }} -U {{ $dbUser | quote }} -d "dbname={{ $dbName }}";
             do echo waiting for database; sleep 2; done;']
           securityContext:
             capabilities:
--- a/charts/kubeclarity/templates/deployment.yaml
+++ b/charts/kubeclarity/templates/deployment.yaml
@@ -40,7 +40,7 @@ spec:
       serviceAccountName: {{ include "kubeclarity.name" . }}
       initContainers:
         - name: '{{ include "kubeclarity.name" . }}-wait-for-pg-db'
-          image: {{ index .Values "kubeclarity-postgresql" "internal" "image" "registry" | default "docker.io" }}/{{ index .Values "kubeclarity-postgresql" "internal" "image" "repository" | default "bitnami/postgresql" }}:{{ index .Values "kubeclarity-postgresql" "internal" "image" "tag" | default "14.4.0-debian-11-r4" }}
+          image: {{ index .Values "kubeclarity-postgresql" "image" "registry" | default "docker.io" }}/{{ index .Values "kubeclarity-postgresql" "image" "repository" | default "bitnami/postgresql" }}:{{ index .Values "kubeclarity-postgresql" "image" "tag" | default "14.4.0-debian-11-r4" }}
           command: ['sh', '-c', 'until pg_isready -h {{ $dbHost }} -p {{ $dbPort }} -U {{ $dbUser }} -d "dbname={{ $dbName }}";
             do echo waiting for database; sleep 2; done;']
           securityContext:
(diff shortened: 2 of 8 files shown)


charts/kubeclarity/values.yaml
```diff
@@ -341,3 +336,15 @@ kubeclarity-postgresql:
...
 #######################################################################################
+
+#######################################################################################
+## KubeClarity Postgres Authentication Values
+kubeclarity-postgresql-auth:
+  # default values can be used if kubeclarity-postgresql is used for pgsql deployment
+  existingSecret: kubeclarity-postgresql-secret
+  database: kubeclarity
+  user: postgres
+  #host: override host if needed
```
sambetts-cisco, line 347: can you move the comment above the field just to make it more obvious, and also add a bit more documentation around it use:
```
# If kubeclarity-postgresql is not enabled then you must provide
# the host of a remote postgresql server.
# host:
```

charts/kubeclarity/values.yaml
```diff
@@ -341,3 +336,15 @@ kubeclarity-postgresql:
 
 # End of KubeClarity Postgres Values
 #######################################################################################
+
+#######################################################################################
+## KubeClarity Postgres Authentication Values
+kubeclarity-postgresql-auth:
```
FrimIdan, line 342: can you explain why you extracted the values from `kubeclarity-postgresql.auth`?

charts/kubeclarity/values.yaml
```diff
@@ -302,11 +302,6 @@ kubeclarity-postgresql:
     # user:
     # password:
 
-  ## Setup database name and password
-  auth:
-    existingSecret: kubeclarity-postgresql-secret
-    database: kubeclarity
```
sambetts-cisco, line 308: Everything under kubeclarity-postgresql are part of the value schema from the postgres sub-chart, these can not be removed from here otherwise kubeclarity-postgresql will not be configured properly.

I think we need to have a new section that isn't reuse for remote postgres, and then template based on kubeclarity-postgresql being enabled or remote being enabled.

sambetts-cisco (changes requested):
Hi @Mesh33, Thanks for the PR, I think this one will need a little more work as its not backwards compatible and is probably breaking the kubeclarity-postgresql install.

As with your other PR, please add more details into the commit message, and same comment about the use of the conventional commit style.

charts/kubeclarity/values.yaml
Mesh33: My idea was that kubeclarity-postgresql was used for the bitnami postgresql instance deployed when `kubeclarity-postgresql.enabled` boolean is true.

I wanted to extract that because `kubeclarity-postgresql-auth` mapping would configure both the external postgresql or the internal one (the one deployed in the helm dependencies).

As there can only be one database, I wanted to mutualize that.

charts/kubeclarity/values.yaml
Mesh33: I'll do that. Thanks for your advice.

charts/kubeclarity/values.yaml
FrimIdan: but now `existingSecret` and `database` can't be configured for the bitnami postgresql instance

charts/kubeclarity/values.yaml
Mesh33: I put back `kubeclarity-postgresql` to normal and added user and port configuration there.

charts/kubeclarity/values.yaml
Mesh33: I added a `kubeclarity-postgresql-external` key to avoid confusion.

charts/kubeclarity/values.yaml
```diff
@@ -341,3 +343,19 @@ kubeclarity-postgresql:
 
 # End of KubeClarity Postgres Values
 #######################################################################################
+
+#######################################################################################
+## KubeClarity External Postgres Values
+
+# notes: self deployed postgresql and external-postgresql can't be used at the same time
+# use this if you wan't to use an external postgresql database as your backend
```
sambetts-cisco, line 351: I think it would be a good idea to add something like this into the template for safety:
```
{{- if .Values.kubeclarity-postgresql.enabled }}
{{- if .Values.kubeclarity-postgresql-external.enabled }}
{{- fail "kubeclarity-postgresql and kubeclarity-postgresql-external can not both be enabled!" }}
{{- end }}
{{- end }}
```

charts/kubeclarity/values.yaml
```diff
@@ -341,3 +343,19 @@ kubeclarity-postgresql:
...
+
+#######################################################################################
+## KubeClarity External Postgres Values
+
+# notes: self deployed postgresql and external-postgresql can't be used at the same time
+# use this if you wan't to use an external postgresql database as your backend 
+kubeclarity-postgresql-external:
+  enabled: false
+  auth:
+    existingSecret: kubeclarity-postgresql-secret
```
sambetts-cisco, line 355: We use the key "existingSecret" for kubeclarity-postgresql because the secret we create for postgres is an existing secret from the sub-chart's perspective.

In the case of kubeclarity-postgresql-external, the secret isn't an existingSecret, we're creating it but that means you'll have to put your external DB password into the values.yaml which isn't ideal.

I think for external we might want two options:
 - use an existing secret (user should create this before doing the helm install so that helm never knows the password, this should be the recommended approach)
 - create a secret, use the template we already have and let the user provide their password in the values.yaml

For this PR if we want to leave the existing behaviour then I think we just need to change the field name to indicate that its not an existing secret but instead will be created.

sambetts-cisco (changes requested):

charts/kubeclarity/values.yaml
Mesh33: I added this at the start of deployment.yaml, do you think this is the right place to put this ? Or should those safety check be elsewhere ?

sambetts-cisco: I think that makes sense for now, perhaps if we have more in the future we'll create a _validation file like the _helpers one

Mesh33:
@sambetts-cisco I did some changes by setting everything under `internal` and `external` key nested under `kubeclarity-postgresql` and also taking your comments into account.

charts/kubeclarity/values.yaml
```diff
@@ -282,13 +281,62 @@ kubeclarity-sbom-db:
 ## KubeClarity Postgres Values
 
 kubeclarity-postgresql:
-  enabled: true
```
sambetts-cisco, line 285: This structure is nice, it means that we can do some common things, I gave me the idea that we could do something like:

```
kubeclarity-postgresql:
   createPasswordSecret: true
   createPasswordSecretName: kubeclarity-postgresql-secret
   # Default is postgres-password
   # createPasswordSecretKey: 

   internal:
      ...
      auth:
        # If createPasswordSecret true, then should match createPasswordSecretName, 
        # otherwise set it to your own secret
        existingSecret: kubeclarity-postgresql-secret
   external:
      ...
      auth:
        # If createPasswordSecret true, then should match createPasswordSecretName
        # otherwise set it to your own secret
        existingSecret: kubeclarity-postgresql-secret
```

if kubeclarity-postgresql.createPasswordSecret == true, then create the secret with the password from global. databasePassword. That way we can control creating the secret separately from using the secret in the internal/external, the benefit here is we get the option to use an existing secret for internal postgres too if someone doesn't want to put their password in helm.

sambetts-cisco (changes requested):
Hey @Mesh33, a nice few updates here, I do like the new structure, and it gave me some ideas (see the comments in values.yaml).

One concern I have and I think I'd like @FrimIdan and/or @akpsgit's opinion on this, any existing user values.yaml using the old structure will not be compatible, so we'd need to ensure we communicate that correctly to avoid users attempting to upgrade and accidentally uninstalling postgres because they have kubeclarity-postgres.enabled true, and not kubeclarity-postgres.internal.enabled true.

charts/kubeclarity/values.yaml
sambetts-cisco: In the future please allow the original commenter to resolve the conversations so they can determine if their comment was resolved or not.

charts/kubeclarity/values.yaml
Mesh33: Thanks for your input.
PR updated with your idea.

FrimIdan:
> Hey @Mesh33, a nice few updates here, I do like the new structure, and it gave me some ideas (see the comments in values.yaml).
> 
> One concern I have and I think I'd like @FrimIdan and/or @akpsgit's opinion on this, any existing user values.yaml using the old structure will not be compatible, so we'd need to ensure we communicate that correctly to avoid users attempting to upgrade and accidentally uninstalling postgres because they have kubeclarity-postgres.enabled true, and not kubeclarity-postgres.internal.enabled true.

@Mesh33 @sambetts-cisco I really don't get why we need to introduce the new `kubeclarity-postgres.internal` section.
I would love to make sure that all changes will be backward compatible, is it a problem?

sambetts-cisco (changes requested):
A few more comments, I need to sync with @FrimIdan to decide on how we want to approach the backwards compatibility issue.

charts/kubeclarity/Chart.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
   - name: postgresql
     version: "11.6.12"
     repository: https://charts.bitnami.com/bitnami
-    condition: kubeclarity-postgresql.enabled
+    condition: kubeclarity-postgresql.internal.enabled
     alias: kubeclarity-postgresql
```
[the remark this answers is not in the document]
sambetts-cisco: This comment is out of date given the last review comment, so resolving

Mesh33:
> @Mesh33 thanks for being patient with us on this PR, I've synced with @FrimIdan on the keys and we've decided that keeping the values.yaml backwards compatible would be the best approach.
> 
> In that case having two postgres keys, a new one for external and the existing on for internal would be best, and then also the "createPasswordSecret" keys can move up into a top level key, resulting in something like:
> 
> ```
> kubeclarity-postgresql:
>     enabled: true
>     ...
>     
> kubeclarity-postgresql-external:
>     enabled: false
>     ...
>     
> kubeclarity-postgresql-secret:
>     create: true
>     createPasswordSecretKey: ...
> ```
> 
> @FrimIdan also pointed out that we can probably drop the `createPasswordSecretName` field because we can read it from the "existingSecret" name in the enabled postgres auth section which will avoid user's needing to sync the two fields together.

@FrimIdan @sambetts-cisco I just made some modifications, fitting your suggestions. I tested both external PostgreSQL instance and the helm chart deployed one. Worked fine.

charts/kubeclarity/values.yaml
```diff
@@ -296,33 +295,20 @@ kubeclarity-sbom-db:
...
-    ##  ConfigMap with scripts to be run at first boot
-    ##  NOTE: This will override initdb.scripts
-    # scriptsConfigMap
-    ##  Secret with scripts to be run at first boot (in case it contains sensitive information)
-    ##  NOTE: This can work along initdbScripts or initdbScriptsConfigMap
-    # scriptsSecret:
-    ## Specify the PostgreSQL username and password to execute the initdb scripts
-    # user:
-    # password:
-
```
FrimIdan, line 321: for the review to be more clear, can you please bring back the fields here and not at the end?

charts/kubeclarity/values.yaml
```diff
@@ -296,33 +295,20 @@ kubeclarity-sbom-db:
...
   ## Setup database name and password
   auth:
-    existingSecret: kubeclarity-postgresql-secret
+    user: postgres
```
FrimIdan, line 305: per https://artifacthub.io/packages/helm/bitnami/postgresql#postgresql-common-parameters field name is `username`

Mesh33: @FrimIdan Updated

charts/kubeclarity/values.yaml
Mesh33: @FrimIdan Updated

Also, comment is removed for postgresql image as I find the template clearer like this (no defaults in the image name in the template).
I can put it back as it was if needed.

charts/kubeclarity/templates/deployment.yaml
```diff
@@ -1,3 +1,21 @@
+{{- if index .Values "kubeclarity-postgresql" "enabled" }}
+{{- if index .Values "kubeclarity-postgresql-external" "enabled" }}
+{{- fail "kubeclarity-postgresql and kubeclarity-postgresql-external can not both be enabled!" }}
+{{- end }}
+{{- end }}
+{{- $secretKey := index .Values "kubeclarity-postgresql-secret" "secretKey" -}}
+{{- $secretName := index .Values "kubeclarity-postgresql" "auth" "existingSecret" -}}
+{{- $dbHost := print (include "kubeclarity.name" .) "-postgresql" -}}
+{{- $dbPort := index .Values "kubeclarity-postgresql" "service" "ports" "postgresql" | quote -}}
+{{- $dbUser := index .Values "kubeclarity-postgresql" "auth" "username" -}}
```
FrimIdan, line 10: nit: in the previous code port was not quoted but username and dbHost was quoted

```
-p 5432 -U "postgres"

"{{ include "kubeclarity.name" . }}-postgresql"
```

Mesh33: @FrimIdan Updated

FrimIdan: Sorry for the nit comments but I really wanted to keep the output the same as before the review in the case of a local postgresql.
I've run `helm template` command to compare it and got
```
286c286
<           command: ['sh', '-c', 'until pg_isready -h kubeclarity-kubeclarity-postgresql -p 5432 -U "postgres" -d "dbname=kubeclarity";
---
>           command: ['sh', '-c', 'until pg_isready -h kubeclarity-kubeclarity-postgresql -p 5432 -U postgres -d "dbname=kubeclarity";
371c371
<               value: "kubeclarity-kubeclarity-postgresql"
---
>               value: kubeclarity-kubeclarity-postgresql
375c375
<               value: "postgres"
---
>               value: postgres
443c443
```

I think that the following will solve it
```
10c10
< {{- $dbUser := index .Values "kubeclarity-postgresql" "auth" "username" -}}
---
> {{- $dbUser := index .Values "kubeclarity-postgresql" "auth" "username" | quote  -}}
133c133
<               value: {{ $dbHost }}
---
>               value: {{ $dbHost | quote  }}
```

Mesh33: @FrimIdan I added quotes in the command to fix the current version.

FrimIdan (approved):
LGTM Thanks for your patience, will let @sambetts-cisco to approve it also

Mesh33:
> LGTM Thanks for your patience, will let @sambetts-cisco to approve it also

@FrimIdan Thank you too for your patience ! I think my next PRs will be WAY more efficient x). I needed a start. I well understood the need to minimize current version changes.

charts/kubeclarity/values.yaml
```diff
@@ -296,16 +296,16 @@ kubeclarity-sbom-db:
...
+## Use kubeclarity-postgresql if you want this chart to deploy a PostgreSQL instance
 kubeclarity-postgresql:
   enabled: true
 
-  ## Specify posgtresql image
-  # image:
-    # registry: docker.io
-    # repository: bitnami/postgresql
-    # tag: 14.4.0-debian-11-r4
+  ## Specify posgtresql image used to check if postgresql DB is ready
```
FrimIdan, line 304: `used to check if postgresql DB is ready` - The command should be removed, it is not only for the init container, it is also used by the posgtresql chart following https://artifacthub.io/packages/helm/bitnami/postgresql#postgresql-common-parameters

sambetts-cisco (approved):
Thanks @Mesh33! LGTM!

PR #279, opened 2022-12-14
Fix issue filtering on ApplicationID with no packages or vuls

sambetts-cisco:
If an Application has no vulnerabilties or packages, when filtering on that ApplicationID through the API, the API would return everything instead of filtering to nothing. This ensures that if you filter on ApplicationID which doesn't match anything, then the API returns no results.

### DIFF
--- a/backend/pkg/database/common.go
+++ b/backend/pkg/database/common.go
@@ -88,7 +88,7 @@ func FilterIsBool(db *gorm.DB, column string, value *bool) *gorm.DB {
 }
 
 func FilterIs(db *gorm.DB, column string, values []string) *gorm.DB {
-	if len(values) == 0 {
+	if values == nil {
 		return db
 	}
 	return db.Where(fmt.Sprintf("%s IN ?", column), values)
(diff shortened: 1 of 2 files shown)


backend/pkg/database/common.go
```diff
@@ -88,7 +88,7 @@ func FilterIsBool(db *gorm.DB, column string, value *bool) *gorm.DB {
 }
 
 func FilterIs(db *gorm.DB, column string, values []string) *gorm.DB {
-	if len(values) == 0 {
+	if values == nil {
 		return db
 	}
 	return db.Where(fmt.Sprintf("%s IN ?", column), values)
```
FrimIdan, line 94: so now we will run it on an empty list and will get nothing? That basically the fix, right?

sambetts-cisco: yeah, we get `SELECT ... WHERE id in []` in the query

akpsgit (approved):

FrimIdan (dismissed):

FrimIdan (approved):

PR #285, opened 2023-01-04
Indicate doneness by closing scan done channel

### LINKED ISSUE #283: Starting a scan on non-existing namespaces makes all future scan attempts result in HTTP 500 Internal Server Error
#### What happened:

I install KubeClarity on Rancher desktop, and it runs without error. I can scan existing namespaces without error. However, when I try to scan a namespace that does not exist, all future start scan attempts fail. The kubeclarity deployment needs to be restarted in order to give reasonable responses again.

I expect the scan query for the non-existing namespace to fail gracefully, and subsequent scan attempts on valid namespace to work.

One erroneous put will make all subsequents puts fail.

```
kubectl port-forward -n kubeclarity svc/kubeclarity-kubeclarity 9999:8080

curl -i 'http://localhost:9999/api/runtime/scan/start' \
  -X 'PUT' \
  -H 'Accept: */*' \
  -H 'content-type: application/json' \
  --data-raw '{"namespaces":["does-not-exist"]}'
```

The following request normally completes correctly, but not if an erroneous namespace has been given:
```
curl -i 'http://localhost:9999/api/runtime/scan/start' \
  -X 'PUT' \
  -H 'Accept: */*' \
  -H 'content-type: application/json' \
  --data-raw {"namespaces":["kubeclarity"]}
```

To make the server respond normally again, it needs to be restarted.

```
2022/12/23 08:42:08 [31;1m/build/backend/pkg/database/scheduler.go:58 [35;1mrecord not found
[0m[33m[3.047ms] [34;1m[rows:0][0m SELECT * FROM "scheduler" ORDER BY "scheduler"."id" LIMIT 1
2022/12/23 08:42:08 Serving kube clarity runtime scan a p is at http://[::]:8888
2022/12/23 08:42:08 Serving kube clarity a p is at http://[::]:8080
time="2022-12-23T08:46:15Z" level=error msg="Failed to send scan config to channel" func="github.com/openclarity/kubeclarity/backend/pkg/rest.(*Server).PutRuntimeScanStart" file="/build/backend/pkg/rest/runtime_scan_controller.go:53"
time="2022-12-23T08:49:20Z" level=error msg="Failed to send scan config to channel" func="github.com/openclarity/kubeclarity/backend/pkg/rest.(*Server).PutRuntimeScanStart" file="/build/backend/pkg/rest/runtime_scan_controller.go:53"
```

#### Anything else we need to know?:

I have also noticed the same behaviour on empty namespaces (such as "default"), but report for non-existing namespace to make the issue easier to reproduce.

#### Environment:
- Kubernetes version: Server Version: v1.24.3+k3s1
- KubeClarity version: v2.9.0 Commit: cf188c2a1e4846d9cbf707c7a217a7642bbe7fe3
- KubeClarity Helm Chart: Installed via `helm template ... --version v2.9.0 ...`
- Cloud provider or hardware configuration: Rancher Desktop 1.7.0
- Others: N/A

sambetts-cisco:
Switch to a safer pattern of creating and closing a channel per scan to indicate doneness instead of using messages on the channel which can be lost if the scan sends a message before the caller is ready to receive it.

Fixes #283


backend/pkg/rest/runtime_scan_controller.go
```diff
@@ -61,8 +61,6 @@ func (s *Server) PutRuntimeScanStart(params operations.PutRuntimeScanStartParams
 func (s *Server) PutRuntimeScanStop(_ operations.PutRuntimeScanStopParams) middleware.Responder {
 	// stop scanner
 	s.runtimeScanner.Clear()
-	// stop listening for scanner results
-	s.runtimeScanner.StopCurrentScan()
```
fishkerez, line 65: why removed? how do we stop a scan now?

sambetts-cisco: The "Clear" function which is used as our stop function everywhere else is now used to stop the scan instead of splitting it across two functions like

fishkerez: Did you check it? not sure how is the scanner notified upon stop. Since he is waiting only for done on startScan func

sambetts-cisco: Yeah, ran and aborted a scan from the UI and it works well. The "done" channel notification now indicates complete in any form, whether aborted, failed or completed, and the scan progress is used to determine what state the scan was in when it is "done".

runtime_scan/pkg/scanner/scanner.go
```diff
@@ -258,35 +258,39 @@ func (s *Scanner) initScan() error {
...
+
 	s.scanConfig = scanConfig
 	log.WithFields(s.logFields).Infof("Start scanning...")
 	s.progress.Status = _types.ScanInit
 	if err := s.initScan(); err != nil {
 		s.progress.SetStatus(_types.ScanInitFailure)
-		return fmt.Errorf("failed to initiate scan: %v", err)
+		// Close done channel to indicate that the scan is no longer running
+		close(done)
+		return done, fmt.Errorf("failed to initiate scan: %v", err)
```
fishkerez, line 275: you are returning a closed channel? is that safe?

sambetts-cisco: yup

backend/pkg/rest/runtime_scan_controller.go
fishkerez: I see it happens via the killSignal

backend/pkg/runtime_scanner/scanner.go
```diff
@@ -138,58 +143,57 @@ func (s *RuntimeScanner) setScanConfig(scanConfig *ScanConfig) {
 func (s *RuntimeScanner) startScan(scanConfig *ScanConfig) error {
 	startTime := time.Now().UTC()
 	s.lock.Lock()
+	if s.stopCurrentScanChan != nil {
+		return fmt.Errorf("found existing cancel channel which may mean a scan is already running, call StopCurrentScan() before calling startScan again to ensure any existing scan is aborted and the cancel channel is cleaned up")
+	}
 	s.lastScanStartTime = startTime
-	stop := s.stopCurrentScanChan
 	s.scannedNamespaces = scanConfig.Namespaces
+	s.stopCurrentScanChan = make(chan struct{})
```
FrimIdan, line 151: I think that it is better to save it locally on
```
stop := s.stopCurrentScanChan
```
since it is protected with a lock

runtime_scan/pkg/orchestrator/orchestrator.go
```diff
@@ -107,7 +108,7 @@ func (o *Orchestrator) Clear() {
 
 	log.Infof("Clearing Orchestrator")
 	o.scanner.Clear()
-	o.scanner = _scanner.CreateScanner(o.config, o.clientset)
+	// o.scanner = _scanner.CreateScanner(o.config, o.clientset)
```
FrimIdan, line 111: remove?

backend/pkg/runtime_scanner/scanner.go
sambetts-cisco: I actually choose not to save it locally because we only ever want to have one scan running, if stopCurrentScanChan is already set and the RuntimeScanner tries to set it (its an internal variable so should only be changed by the RuntimeScanner) then it means there is already a scan running and the StopCurrentScan function should be called to clear it.

I was trying to decide whether the startScan function should hold a read lock the entire time its waiting on s.stopCurrentScanChan to prevent a writer from replacing it behind its back

runtime_scan/pkg/orchestrator/orchestrator.go
sambetts-cisco: done

runtime_scan/pkg/orchestrator/orchestrator.go
```diff
@@ -85,12 +85,13 @@ func (o *Orchestrator) Stop() {
...
 }
 
-func (o *Orchestrator) Scan(scanConfig *_config.ScanConfig, scanDone chan struct{}) error {
-	if err := o.getScanner().Scan(scanConfig, scanDone); err != nil {
-		return fmt.Errorf("failed to scan: %v", err)
+func (o *Orchestrator) Scan(scanConfig *_config.ScanConfig) (chan struct{}, error) {
+	o.scanner = _scanner.CreateScanner(o.config, o.clientset)
+	done, err := o.getScanner().Scan(scanConfig)
+	if err != nil {
+		return done, fmt.Errorf("failed to scan: %v", err)
```
akpsgit, line 92: maybe better to return nil here instead of done?
Not sure it's clear what to do when you get both an error and a channel

runtime_scan/pkg/scanner/scanner.go
akpsgit: same

backend/pkg/runtime_scanner/scanner.go
FrimIdan: I know that we only have one scan running, but since we closing and replacing the channel under a lock I think it will be safer to save it locally here to avoid holding a read lock the entire time it's waiting on it.

sambetts-cisco: updated it so that we've got a nice function for allocating a new stop channel that should always be used to write stopCurrentScanChan and it return the allocated channel, then I can do a defer on the StopCurrentChan for cleanup when the scan is completed.

runtime_scan/pkg/orchestrator/orchestrator.go
sambetts-cisco: updated it so that it returns nil whenever there is an error to be returned and moved the channel setup later in the functions that return an error:
https://github.com/openclarity/kubeclarity/compare/750c55cd399c68f262348dd7dca6b5c51d289f2f..994d85defae90bcfd22cb5c165ada99ddfa8fd63

FrimIdan (commented):
LGTM

FrimIdan (approved):
LGTM

PR #291, opened 2023-01-10
Fix trivy analyzer detects only os packages

boris257:
Trivy analyzer should also detect programming language dependencies.

### DIFF
--- a/shared/pkg/analyzer/trivy/trivy.go
+++ b/shared/pkg/analyzer/trivy/trivy.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/aquasecurity/trivy/pkg/commands/artifact"
 	trivyFlag "github.com/aquasecurity/trivy/pkg/flag"
+	trivyTypes "github.com/aquasecurity/trivy/pkg/types"
 	trivyUtils "github.com/aquasecurity/trivy/pkg/utils"
 
 	"github.com/openclarity/kubeclarity/shared/pkg/analyzer"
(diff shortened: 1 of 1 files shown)


sambetts-cisco:
Hi @boris257 thanks for the PR, have you confirmed that this fix works? Looking through the Trivy code I can only see the Vulnerability Options getting read if the Vulnerability security check is enabled (which we don't enable in the analyzer because we just want the SBOM) https://github.com/aquasecurity/trivy/blob/c110c4e028d50a2253c1d18432b075c5f83a0d34/pkg/scanner/local/scan.go#L218.

My understanding is the ListAllPkgs option is supposed to list all the Language and OS package regardless of what is configured in the Vulnerability option.

boris257:
Hi! Yes, I can confirm that this fix works. The reason is the language package analyzer is disabled by default: https://github.com/aquasecurity/trivy/blob/c110c4e028d50a2253c1d18432b075c5f83a0d34/pkg/commands/artifact/run.go#L445-L448.

> My understanding is the ListAllPkgs option is supposed to list all the Language and OS package regardless of what is configured in the Vulnerability option.

As a counterexample, you can run 
```
trivy image --list-all-pkgs --vuln-type os --format cyclonedx your-image-ref  
trivy image --list-all-pkgs --vuln-type os,library --format cyclonedx your-image-ref
```
and see the difference.

shared/pkg/analyzer/trivy/trivy.go
```diff
@@ -96,6 +97,9 @@ func (a *Analyzer) Run(sourceType utils.SourceType, userInput string) error {
 				Output:       &output,     // Save the output to our local buffer instead of Stdout
 				ListAllPkgs:  true,        // By default Trivy only includes packages with vulnerabilities, for full SBOM set true.
 			},
+			VulnerabilityOptions: trivyFlag.VulnerabilityOptions{
+				VulnType: trivyTypes.VulnTypes, // Scan all vuln types trivy supports
```
sambetts-cisco, line 101: Could we update the comment here to explain that even though we don't have the vuln security check turned on we need to provide the whole list otherwise trivy disables the analyzer for language packages

sambetts-cisco (changes requested):
Great spot! I couldn't for the life of me find that line in the trivy code. I think it would be nice if we could update the comment so I've left a review. Could you also add a bit of an explanation into the commit message so that we have it when we look in the git log for why a line was changed.

shared/pkg/analyzer/trivy/trivy.go
boris257: @sambetts-cisco, yes, done.

sambetts-cisco (approved):

PR #294, opened 2023-01-11
Add support for exposing Prometheus metrics endpoint

### LINKED ISSUE #63: Metrics
Hi,
Does kubei expose any `/metrics` with equivalent info that the web exposes (about vulnerabilities found)?

Was thinking about monitoring and alerting on vulnerabilities found, for example via Prometheus.

nostra:
This PR will add a /metrics endpoint to KubeClarity. It needs to be enabled with a config value, with a default of not being enabled. 

In order to expose metrics from KubeClarity to Prometheus, you need to enable
the endpoint by setting the environment variable `PROMETHEUS_REFRESH_INTERVAL_SECONDS`
to value larger than zero, for instance 300 which means that the Prometheus endpoint will get updated values every 5 minutes.

KubeClarity uses healthz for readiness / liveness, and as the metrics is added with `http.Handle("/metrics", promhttp.Handler())`, this will make it live on port 8081.

I have added an Alertmanager example, and a Grafana dashboard.

Grafana Dashboard

This PR will close #63

### DIFF
--- a/backend/pkg/backend/backend.go
+++ b/backend/pkg/backend/backend.go
@@ -19,14 +19,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/Portshift/go-utils/healthz"
+	k8sutils "github.com/Portshift/go-utils/k8s"
+	"github.com/openclarity/kubeclarity/backend/pkg/metrics"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/client-go/kubernetes"
 	"os"
 	"os/signal"
 	"syscall"
-
-	"github.com/Portshift/go-utils/healthz"
-	k8sutils "github.com/Portshift/go-utils/k8s"
-	log "github.com/sirupsen/logrus"
-	"k8s.io/client-go/kubernetes"
 
 	_config "github.com/openclarity/kubeclarity/backend/pkg/config"
 	_database "github.com/openclarity/kubeclarity/backend/pkg/database"
@@ -111,6 +111,10 @@ func Run() {
 	}
 	restServer.Start(errChan)
 	defer restServer.Stop()
+
+	if config.PrometheusRefreshIntervalSeconds > 0 {
+		metrics.ProduceMetrics(dbHandler, config.PrometheusRefreshIntervalSeconds).StartRecordingMetrics()
+	}
 
 	healthServer.SetIsReady(true)
 	log.Info("KubeClarity backend is ready")
--- a/backend/pkg/backend/backend.go
+++ b/backend/pkg/backend/backend.go
@@ -19,14 +19,16 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/Portshift/go-utils/healthz"
-	k8sutils "github.com/Portshift/go-utils/k8s"
-	"github.com/openclarity/kubeclarity/backend/pkg/metrics"
-	log "github.com/sirupsen/logrus"
-	"k8s.io/client-go/kubernetes"
 	"os"
 	"os/signal"
 	"syscall"
+
+	"github.com/openclarity/kubeclarity/backend/pkg/metrics"
+
+	"github.com/Portshift/go-utils/healthz"
+	k8sutils "github.com/Portshift/go-utils/k8s"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/client-go/kubernetes"
 
 	_config "github.com/openclarity/kubeclarity/backend/pkg/config"
 	_database "github.com/openclarity/kubeclarity/backend/pkg/database"
--- a/backend/pkg/backend/backend.go
+++ b/backend/pkg/backend/backend.go
@@ -75,8 +75,7 @@ func Run() {
 
 	healthServer.SetIsReady(false)
 
-	// globalCtx, globalCancel := context.WithCancel(context.Background())
-	_, globalCancel := context.WithCancel(context.Background())
+	globalCtx, globalCancel := context.WithCancel(context.Background())
 	defer globalCancel()
 
 	log.Info("KubeClarity backend is running")
@@ -113,7 +112,7 @@ func Run() {
 	defer restServer.Stop()
 
 	if config.PrometheusRefreshIntervalSeconds > 0 {
-		metrics.CreateMetrics(dbHandler, config.PrometheusRefreshIntervalSeconds).StartRecordingMetrics(context.Background())
+		metrics.CreateMetrics(dbHandler, config.PrometheusRefreshIntervalSeconds).StartRecordingMetrics(globalCtx)
 	}
 
 	healthServer.SetIsReady(true)
--- a/backend/pkg/backend/backend.go
+++ b/backend/pkg/backend/backend.go
@@ -113,7 +113,7 @@ func Run() {
 	defer restServer.Stop()
 
 	if config.PrometheusRefreshIntervalSeconds > 0 {
-		metrics.ProduceMetrics(dbHandler, config.PrometheusRefreshIntervalSeconds).StartRecordingMetrics()
+		metrics.CreateMetrics(dbHandler, config.PrometheusRefreshIntervalSeconds).StartRecordingMetrics(context.Background())
 	}
 
 	healthServer.SetIsReady(true)
--- a/backend/pkg/metrics/README.md
+++ b/backend/pkg/metrics/README.md
@@ -5,7 +5,14 @@ the endpoint by setting the environment variable `PROMETHEUS_REFRESH_INTERVAL_SE
 to value larger than zero. There is no particular need to scrape these metrics very
 often, so once every 300 seconds is fine.
 
-You can set this value in the `values.yaml` file.
+Enable Prometheus by flipping the "enable" flag to "true" in the `values.yaml`:
+
+```
+kubeclarity:
+  prometheus:
+    enabled: false
+    refreshIntervalSeconds: 300 
+```
 
 The metrics that are exposed are:
 
--- a/backend/pkg/metrics/README.md
+++ b/backend/pkg/metrics/README.md
@@ -9,16 +9,17 @@ You can set this value in the `values.yaml` file.
 
 The metrics that are exposed are:
 
-| Metric                                              | Explanation                                        |
-|-----------------------------------------------------|----------------------------------------------------|
-| kubeclarity_application_vulnerability               | Count of vulnerabilities per application and level |
-| kubeclarity_number_of_applications                  | The total number of applications                   |
-| kubeclarity_number_of_fixable_vulnerabilities       | The number of fixable vulnerabilities per severity |
-| kubeclarity_number_of_fixable_vulnerabilities_total | The total number of fixable vulnerabilities        |
-| kubeclarity_number_of_packages                      | The total number of packages                       |
-| kubeclarity_number_of_resources                     | The total number of resources                      |
-| kubeclarity_number_of_vulnerabilities               | The number of vulnerabilities per severity         |
-| kubeclarity_number_of_vulnerabilities_total         | The total number of vulnerabilities                |
+| Metric                                              | Explanation                                                        |
+|-----------------------------------------------------|--------------------------------------------------------------------|
+| kubeclarity_application_vulnerability               | Count of vulnerabilities per application, environment and severity |
+| kubeclarity_number_of_applications                  | The total number of applications                                   |
+| kubeclarity_number_of_fixable_vulnerabilities       | The number of fixable vulnerabilities per severity                 |
+| kubeclarity_number_of_fixable_vulnerabilities_total | The total number of fixable vulnerabilities                        |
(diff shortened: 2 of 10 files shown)


backend/pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+func ProduceMetrics(dbHandler *database.Handler, refreshInterval int) *Server {
+	return &Server{
+		refreshInterval: refreshInterval,
+		dbHandler:       dbHandler,
+	}
+}
+
+func init() {
+	logrus.Infof("Adding /metrics endpoint to http server assumed started with healthz")
+	http.Handle("/metrics", promhttp.Handler())
```
FrimIdan, line 75: would it make sense to avoid using `init()` and move the code under `ProduceMetrics`?

backend/pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+		Help: "Count of vulnerabilities per application and level",
+	}, []string{"name", "environment", "severity"})
+)
+
+type Server struct {
+	refreshInterval int
+	dbHandler       *database.Handler
+}
+
+func ProduceMetrics(dbHandler *database.Handler, refreshInterval int) *Server {
```
FrimIdan, line 66: `ProduceMetrics` - why choosing that name and not a basic `New`/`Create`?

backend/pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	http.Handle("/metrics", promhttp.Handler())
+}
+
+func (s *Server) recordSummaryCounters() {
+	pkgCount, _ := s.dbHandler.PackageTable().Count(nil)
+	packageCounter.Set(float64(pkgCount))
+	appCount, _ := s.dbHandler.ApplicationTable().Count(nil)
+	applicationCounter.Set(float64(appCount))
+	resCount, _ := s.dbHandler.ResourceTable().Count(nil)
+	resourceCounter.Set(float64(resCount))
```
FrimIdan, line 84: why ignoring the errors here?

backend/pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (s *Server) recordTrendCounters() {
+	trends, err := s.dbHandler.NewVulnerabilityTable().
+		GetNewVulnerabilitiesTrends(operations.GetDashboardTrendsVulnerabilitiesParams{
+			StartTime: strfmt.DateTime(time.Now().Add(-60 * time.Minute)),
```
FrimIdan, line 90: `-60` - should it be the refresh time?

backend/pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+		Help: "The number of fixable vulnerabilities per severity",
+	}, []string{"severity"})
+	vulnerability = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: Prefix + "_number_of_vulnerabilities",
+		Help: "The number of vulnerabilities per severity",
+	}, []string{"severity"})
+	trendGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: Prefix + "_vulnerability_trend",
+		Help: "Incoming vulnerabilities within the last hour",
+	}, []string{"severity"})
```
FrimIdan, line 54: Missing a documentation in the README.md.
Also, not sure about the description here, what is the meaning of `Incoming vulnerabilities`?
trends are used in the dashboard to show new vulnerabilities (to that system) that was found in a time period.

backend/pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	for _, trend := range trends {
+		for _, vul := range trend.NumOfVuls {
+			trendGauge.WithLabelValues(string(vul.Severity)).Set(float64(vul.Count))
+		}
+	}
+}
+
+// Gauges for fixable vulnerabilities
+func (s *Server) recordFixableVulnerability() {
+	fixableVulnerabilityCount, _ := s.dbHandler.VulnerabilityTable().CountVulnerabilitiesWithFix()
```
FrimIdan, line 106: again, not sure why ignoring the returned error

backend/pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	var total uint32 = 0
+	var totalWithFix uint32 = 0
+	for _, fix := range fixableVulnerabilityCount {
+		total += fix.CountTotal
+		totalWithFix += fix.CountWithFix
+		fixableVulnerability.WithLabelValues(fmt.Sprintf("%s", fix.Severity)).Set(float64(fix.CountWithFix))
```
FrimIdan, line 112: `fmt.Sprintf("%s", fix.Severity)` - why not `string(vul.Severity)` as before?

backend/pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+		Name: Prefix + "_number_of_vulnerabilities",
+		Help: "The number of vulnerabilities per severity",
+	}, []string{"severity"})
+	trendGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: Prefix + "_vulnerability_trend",
+		Help: "Incoming vulnerabilities within the last hour",
+	}, []string{"severity"})
+	applicationGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: Prefix + "_application_vulnerability",
+		Help: "Count of vulnerabilities per application and level",
```
FrimIdan, line 57: maybe better to note here that it only list application with environment?

charts/kubeclarity/values.yaml
```diff
@@ -37,13 +37,16 @@ kubeclarity:
 
   enableDBInfoLog: false
 
-  podAnnotations: {}
+  # A value greater than 0 will enable the Prometheus metrics endpoint:
+  prometheusRefreshIntervalSeconds: 0
```
FrimIdan, line 41: I think it will be better to create section for `prometheus` (with an enable bit) if we will want in the future to add more values under

```
prometheus:
  enabled: false
  refreshIntervalSeconds: 300
```

I think it is more clear how to enable/disable it. WDYT?

FrimIdan (changes requested):
Thanks for the contribution it looks great!
I've added some small comments.

backend/pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,158 @@
...
+				applicationGauge.WithLabelValues(application.Name, env, "HIGH").Set(float64(application.TotalHighCount))
+				applicationGauge.WithLabelValues(application.Name, env, "CRITICAL").Set(float64(application.TotalCriticalCount))
+			}
+		}
+	}
+}
+
+func (s *Server) StartRecordingMetrics() {
+	go func() {
+		for {
```
fishkerez, line 149: Can you pass a context here, and do something like this:
```
for {
			select {
			case <-ctx.done:
				log.Info("received stop event")
				return
			case time.after(refreshInterval):
				....
			}
		}
```

nostra:
Local trouble made me drop the fork. I'll get it recreated shortly.

backend/pkg/metrics/prometheus_metrics.go
FrimIdan: I think a warning print and continue would be good.

backend/pkg/backend/backend.go
```diff
@@ -113,7 +113,7 @@ func Run() {
 	defer restServer.Stop()
 
 	if config.PrometheusRefreshIntervalSeconds > 0 {
-		metrics.ProduceMetrics(dbHandler, config.PrometheusRefreshIntervalSeconds).StartRecordingMetrics()
+		metrics.CreateMetrics(dbHandler, config.PrometheusRefreshIntervalSeconds).StartRecordingMetrics(context.Background())
```
FrimIdan, line 116: you can use the `globalCtx` that currently is not being used
```
_, globalCancel := context.WithCancel(context.Background())
```

backend/pkg/backend/backend.go
```diff
@@ -19,14 +19,14 @@ import (
...
-	"syscall"
-
 	"github.com/Portshift/go-utils/healthz"
 	k8sutils "github.com/Portshift/go-utils/k8s"
+	"github.com/openclarity/kubeclarity/backend/pkg/metrics"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/client-go/kubernetes"
+	"os"
+	"os/signal"
+	"syscall"
```
FrimIdan, line 29: gofmt/goimport is needed, probably the linter will be angry :)

backend/pkg/metrics/README.md
```diff
@@ -0,0 +1,32 @@
+# Prometheus
+
+In order to expose metrics from KubeClarity to Prometheus, you need to enable
+the endpoint by setting the environment variable `PROMETHEUS_REFRESH_INTERVAL_SECONDS`
+to value larger than zero. There is no particular need to scrape these metrics very
+often, so once every 300 seconds is fine.
+
+You can set this value in the `values.yaml` file.
```
FrimIdan, line 8: Need to mention that there is a need to also enable the Prometheus section in `values.yaml`, right?

backend/pkg/backend/backend.go
nostra: If we get a different result, then there is some strangeness afoot. I try to move the "os" part back to the former place, and do:

```
❯ git diff
diff --git a/backend/pkg/backend/backend.go b/backend/pkg/backend/backend.go
index a5f21a1..b787bc0 100644
--- a/backend/pkg/backend/backend.go
+++ b/backend/pkg/backend/backend.go
@@ -19,14 +19,14 @@ import (
        "context"
        "encoding/json"
        "fmt"
+       "os"
+       "os/signal"
+       "syscall"
        "github.com/Portshift/go-utils/healthz"
        k8sutils "github.com/Portshift/go-utils/k8s"
        "github.com/openclarity/kubeclarity/backend/pkg/metrics"
        log "github.com/sirupsen/logrus"
        "k8s.io/client-go/kubernetes"
-       "os"
-       "os/signal"
-       "syscall"

        _config "github.com/openclarity/kubeclarity/backend/pkg/config"
        _database "github.com/openclarity/kubeclarity/backend/pkg/database"

~/git/fork/kubeclarity expose_prometheus_metrics_endpoint*
❯ go fmt backend/pkg/backend/backend.go
backend/pkg/backend/backend.go

~/git/fork/kubeclarity expose_prometheus_metrics_endpoint
❯ git diff

~/git/fork/kubeclarity expose_prometheus_metrics_endpoint
```

It could be that my platform and timezone is a factor. I sit in CET, on a Mac. I tried to do `make fix` and it exited with
```
can't run linter goanalysis_metalinter: goimports: can't extract issues from gofmt diff output 
... 
can't parse patch: parsing time "2023-01-16 14:26:56" as "2006-01-02 15:04:05 -0700": cannot parse "" as "-0700"

Comment (sambetts-cisco) [reply backend/pkg/backend/backend.go:29]: I've hit the diff issue before, the problem is with the native version of `diff` that mac ships with, they changed it recently so you now have to install the GNU version of `diff` otherwise it can't parse the diff output:
```
brew install diffutils
```

Comment (nostra) [reply backend/pkg/backend/backend.go]: Thanks! I would have had a hard time figuring this out myself. `make fix` now works, and I expect that to make the linter happy.

**Review by FrimIdan — DISMISSED** [review]
Thank you!

File: charts/kubeclarity/templates/deployment.yaml [hunk 14]
```diff
@@ -100,6 +100,10 @@ spec:
           env:
             - name: ENABLE_DB_INFO_LOGS
               value: "{{ .Values.kubeclarity.enableDBInfoLog }}"
+{{- if index .Values "prometheus" "enabled" }}
```
FrimIdan, line 103: typo - it should be `if .Values.kubeclarity.prometheus.enabled`

FrimIdan (dismissed):

backend/pkg/metrics/prometheus_metrics.go
```diff
@@ -0,0 +1,191 @@
...
+				applicationGauge.WithLabelValues(application.Name, env, "HIGH").Set(float64(application.TotalHighCount))
+				applicationGauge.WithLabelValues(application.Name, env, "CRITICAL").Set(float64(application.TotalCriticalCount))
+			}
+		}
+	}
+}
+
+func (s *Server) StartRecordingMetrics(ctx context.Context) {
+	go func() {
+		for {
```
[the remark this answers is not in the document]
FrimIdan: Sorry, I missed that "fix" - for sure you need to run the for loop in a goroutine

FrimIdan (approved):

PR #303, opened 2023-01-17
use installed namespace for CREDS_SECRET_NAMESPACE

MasayaAoyama:
Hi

we cannot use cluster credentials for Google Artifact Registry, etc, when we install kubeclarity by helm to our namespace except kubeclarity.
we have two options;

1. create kubeclarity namespace, and then create cluster credentials to the namespace.

2. set CREDS_SECRET_NAMESPACE with installed namespace at generating manifests by helm

I think option 2 is better.
If you think, please merge this :)

And if you want, I will add documentation at following section in README.md.
https://github.com/openclarity/kubeclarity#private-registries-support-for-k8s-runtime-scan


FrimIdan (approved):

FrimIdan (changes requested):
Thanks for the contribution, please update the README as suggested and we can push it.

MasayaAoyama:
@FrimIdan 
sorry for late.
I added documentations for CREDS_SECRET_NAMESPACE.

README.md
```diff
@@ -358,6 +358,8 @@ kubeclarity scan registry/nginx:private --config $HOME/own-kubeclarity-config
 Kubeclarity is using [k8schain](https://github.com/google/go-containerregistry/tree/main/pkg/authn/k8schain#k8schain) of google/go-containerregistry for authenticating to the registries.
 If the necessary service credentials are not discoverable by the k8schain, they can be defined via secrets described below.
 
+In addition, if service credentials are not located in "kubeclarity" Namespace, please set CREDS_SECRET_NAMESPACE to kubeclarity Deployment.
```
FrimIdan, line 361: Since the helm chart is setting `CREDS_SECRET_NAMESPACE` to the release namespace, I think that the comment should be in general for the creation of the secrets to be created in the release namespace and not in `kubeclarity` namespace as described in the examples below.

MasayaAoyama: @FrimIdan Thank you for review. I appended case of using helm charts.

FrimIdan (approved):

PR #308, opened 2023-01-19
fix: allow setting proxies when trivy is disabled

### LINKED ISSUE #307: [chart] templating fails when setting proxies
#### What happened:
When setting `kubeclarity-runtime-scan.httpsProxy` and/or `kubeclarity-runtime-scan.httpProxy`  in `values.yaml` templating fails with the following error:

```
Error: template: kubeclarity/templates/scanner-template-configmap.yaml:112:356: executing "kubeclarity/templates/scanner-template-configmap.yaml" at <include "trivy.fullname" (index .Subcharts "kubeclarity-trivy-server")>: error calling include: template: no template "trivy.fullname" associated with template "gotpl"

Use --debug flag to render out invalid YAML
```

Helm templating succeeds with proxy settings defined, without the need to enable trivy.

```
helm fetch kubeclarity/kubeclarity --untar
cd kubeclarity
helm template . --set kubeclarity-runtime-scan.httpsProxy="https://proxyhost.tld:port"
or
helm template . --set kubeclarity-runtime-scan.httpProxy="http://proxyhost.tld:port"
```
Both generate the error:
```
Error: template: kubeclarity/templates/scanner-template-configmap.yaml:112:356: executing "kubeclarity/templates/scanner-template-configmap.yaml" at <include "trivy.fullname" (index .Subcharts "kubeclarity-trivy-server")>: error calling include: template: no template "trivy.fullname" associated with template "gotpl"

Use --debug flag to render out invalid YAML
```

(e.g. `kubectl logs -n kubeclarity --selector=app=kubeclarity`)

#### Environment:
- Kubernetes version (use `kubectl version --short`): any
- KubeClarity version (use `kubectl -n kubeclarity exec deploy/kubeclarity -- ./backend version`) v2.11.0
- KubeClarity Helm Chart version (use `helm -n kubeclarity list`) v2.11.0 (olders versions also affected)
- helm version: version.BuildInfo{Version:"v3.11.0", GitCommit:"472c5736ab01133de504a826bd9ee12cbe4e7904", GitTreeState:"clean", GoVersion:"go1.19.5"}

florentinadolf:
Should fix #307

### DIFF
--- a/charts/kubeclarity/templates/scanner-template-configmap.yaml
+++ b/charts/kubeclarity/templates/scanner-template-configmap.yaml
@@ -1,3 +1,10 @@
+{{- $noproxy := list (print (include "kubeclarity.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-runtime-scan" "resultServicePort")) -}}
+{{- if index .Values "kubeclarity-grype-server" "enabled" -}}
+{{- $noproxy = append $noproxy (print (include "kubeclarity.grype-server.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-grype-server" "servicePort")) -}}
+{{- end -}}
+{{- if index .Values "kubeclarity-trivy-server" "enabled" -}}
+{{- $noproxy = append $noproxy (print ((index .Values "kubeclarity-trivy-server" "service" "name") | default (include "trivy.fullname" (index .Subcharts "kubeclarity-trivy-server"))) "." .Release.Namespace ":" (index .Values "kubeclarity-trivy-server" "service" "port")) -}}
+{{- end -}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -99,13 +106,6 @@ data:
             - name: SCANNER_DEPENDENCY_TRACK_API_KEY
               value: {{ index .Values "kubeclarity-runtime-scan" "vulnerability-scanner" "scanner" "dependency-track" "apiKey" }}
 {{- end }}
-{{- $noproxy := list (print (include "kubeclarity.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-runtime-scan" "resultServicePort")) -}}
-{{- if index .Values "kubeclarity-grype-server" "enabled" -}}
-{{- $noproxy = append $noproxy (print (include "kubeclarity.grype-server.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-grype-server" "servicePort")) -}}
-{{- end -}}
-{{- if index .Values "kubeclarity-trivy-server" "enabled" -}}
-{{- $noproxy = append $noproxy (print ((index .Values "kubeclarity-trivy-server" "service" "name") | default (include "trivy.fullname" (index .Subcharts "kubeclarity-trivy-server"))) "." .Release.Namespace ":" (index .Values "kubeclarity-trivy-server" "service" "port")) -}}
-{{- end -}}
 {{- if index .Values "kubeclarity-runtime-scan" "httpsProxy" }}
             - name: HTTPS_PROXY
               value: {{ index .Values "kubeclarity-runtime-scan" "httpsProxy" }}
--- a/charts/kubeclarity/templates/scanner-template-configmap.yaml
+++ b/charts/kubeclarity/templates/scanner-template-configmap.yaml
@@ -99,6 +99,13 @@ data:
             - name: SCANNER_DEPENDENCY_TRACK_API_KEY
               value: {{ index .Values "kubeclarity-runtime-scan" "vulnerability-scanner" "scanner" "dependency-track" "apiKey" }}
 {{- end }}
+{{- $noproxy := list (print (include "kubeclarity.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-runtime-scan" "resultServicePort")) -}}
+{{- if index .Values "kubeclarity-grype-server" "enabled" -}}
+{{- $noproxy = append $noproxy (print (include "kubeclarity.grype-server.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-grype-server" "servicePort")) -}}
+{{- end -}}
+{{- if index .Values "kubeclarity-trivy-server" "enabled" -}}
+{{- $noproxy = append $noproxy (print ((index .Values "kubeclarity-trivy-server" "service" "name") | default (include "trivy.fullname" (index .Subcharts "kubeclarity-trivy-server"))) "." .Release.Namespace ":" (index .Values "kubeclarity-trivy-server" "service" "port")) -}}
+{{- end -}}
 {{- if index .Values "kubeclarity-runtime-scan" "httpsProxy" }}
             - name: HTTPS_PROXY
               value: {{ index .Values "kubeclarity-runtime-scan" "httpsProxy" }}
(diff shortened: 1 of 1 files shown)


charts/kubeclarity/templates/scanner-template-configmap.yaml
```diff
@@ -109,10 +109,18 @@ data:
 {{- end}}
 {{- if index .Values "kubeclarity-runtime-scan" "httpsProxy" }}
             - name: NO_PROXY
+{{- if index .Values "kubeclarity-trivy-server" "enabled" }}
               value: {{ include "kubeclarity.name" . }}.{{ .Release.Namespace }}:{{ index .Values "kubeclarity-runtime-scan" "resultServicePort" }},{{ include "kubeclarity.grype-server.name" . }}.{{ .Release.Namespace }}:{{ index .Values "kubeclarity-grype-server" "servicePort" }},{{ (index .Values "kubeclarity-trivy-server" "service" "name") | default (include "trivy.fullname" (index .Subcharts "kubeclarity-trivy-server")) }}.{{ .Release.Namespace }}:{{ index .Values "kubeclarity-trivy-server" "service" "port" }}
+{{- else}}
+              value: {{ include "kubeclarity.name" . }}.{{ .Release.Namespace }}:{{ index .Values "kubeclarity-runtime-scan" "resultServicePort" }},{{ include "kubeclarity.grype-server.name" . }}.{{ .Release.Namespace }}:{{ index .Values "kubeclarity-grype-server" "servicePort" }}
```
sambetts-cisco, line 115: I think we should approach this a little differently to allow it to be extensible in the future when/if we add additional server types for other plugins. We can use some of the tools in the helm template to contruct a list in a variable at the top of this file like:
```
{{ $noproxy := list (print (include "kubeclarity.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-runtime-scan" "resultServicePort")) }}
{{ if index .Values "kubeclarity-grype-server" "enabled" }}
{{ $noproxy = append $noproxy (print (include "kubeclarity.grype-server.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-grype-server" "servicePort")) }}
{{ end }}
{{ if index .Values "kubeclarity-trivy-server" "enabled" }}
{{ $noproxy = append $noproxy (print ((index .Values "kubeclarity-trivy-server" "service" "name") | default (include "trivy.fullname" (index .Subcharts "kubeclarity-trivy-server"))) "." .Release.Namespace ":" (index .Values "kubeclarity-trivy-server" "service" "port")) }}
{{ end}}
```

and then here the template will become something like:

```
{{- if index .Values "kubeclarity-runtime-scan" "httpsProxy" }}
- name: NO_PROXY
   value: {{ join "," $noproxy }}
{{- end}}
```

it also means that we can avoid duplicating the logic like we do today.

sambetts-cisco (changes requested):
Hi @florentinadolf, thanks for bug report and the PR! Sorry for the regression when we introduced the Trivy server option. I've left a comment about tweaking the approach for this fix to make it easier to avoid this issue in the future.

charts/kubeclarity/templates/scanner-template-configmap.yaml
florentinadolf: Hello @sambetts-cisco i added a function to generate the noProxy hosts to the helper template, hope that's how you expected me to do it.

charts/kubeclarity/templates/_helpers.tpl
```diff
@@ -77,3 +77,17 @@ Sets extra Kubeclarity server Service annotations
...
+Sets extra Kubeclarity server Service annotations
+*/}}
+{{- define "kubeclarity.noproxy" -}}
+{{- $noproxy := list (print (include "kubeclarity.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-runtime-scan" "resultServicePort")) -}}
+{{- if index .Values "kubeclarity-grype-server" "enabled" -}}
+{{- $noproxy = append $noproxy (print (include "kubeclarity.grype-server.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-grype-server" "servicePort")) -}}
+{{- end -}}
+{{- if index .Values "kubeclarity-trivy-server" "enabled" -}}
+{{- $noproxy = append $noproxy (print ((index .Values "kubeclarity-trivy-server" "service" "name") | default (include "trivy.fullname" (index .Subcharts "kubeclarity-trivy-server"))) "." .Release.Namespace ":" (index .Values "kubeclarity-trivy-server" "service" "port")) -}}
+{{- end}}
```
sambetts-cisco, line 91: This variable will get recreated every time this sub template is included, if we define it outside of "kubeclarity.noproxy" then it'll only get created once the first time the templates are loaded. As the data is static I don't think we need to regenerate the list each time. I think we could re-write this like:

```
{{- $noproxy := list (print (include "kubeclarity.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-runtime-scan" "resultServicePort")) -}}
{{- if index .Values "kubeclarity-grype-server" "enabled" -}}
{{- $noproxy = append $noproxy (print (include "kubeclarity.grype-server.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-grype-server" "servicePort")) -}}
{{- end -}}
{{- if index .Values "kubeclarity-trivy-server" "enabled" -}}
{{- $noproxy = append $noproxy (print ((index .Values "kubeclarity-trivy-server" "service" "name") | default (include "trivy.fullname" (index .Subcharts "kubeclarity-trivy-server"))) "." .Release.Namespace ":" (index .Values "kubeclarity-trivy-server" "service" "port")) -}}
{{- end}}

{{- define "kubeclarity.noproxy" -}}
{{ join "," $noproxy }}
{{- end -}}
```

and then have in the main template:

```
- name: NO_PROXY
   value: {{ include "kubeclarity.noproxy" . }}
```

or even drop the sub-template and just do this:

```
- name: NO_PROXY
   value: {{ join "," $noproxy }}
```

sambetts-cisco (changes requested):

charts/kubeclarity/templates/_helpers.tpl
florentinadolf: @sambetts-cisco i removed the helper function and added the parsing of the $noproxy to the scanner-template-configmap it should now only be evaluated once for all 4 instances it is being used.

charts/kubeclarity/templates/scanner-template-configmap.yaml
florentinadolf: $noproxy is now evaluated without helper function.

charts/kubeclarity/templates/scanner-template-configmap.yaml
```diff
@@ -99,6 +99,13 @@ data:
             - name: SCANNER_DEPENDENCY_TRACK_API_KEY
               value: {{ index .Values "kubeclarity-runtime-scan" "vulnerability-scanner" "scanner" "dependency-track" "apiKey" }}
 {{- end }}
+{{- $noproxy := list (print (include "kubeclarity.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-runtime-scan" "resultServicePort")) -}}
+{{- if index .Values "kubeclarity-grype-server" "enabled" -}}
+{{- $noproxy = append $noproxy (print (include "kubeclarity.grype-server.name" .) "." .Release.Namespace ":" (index .Values "kubeclarity-grype-server" "servicePort")) -}}
+{{- end -}}
+{{- if index .Values "kubeclarity-trivy-server" "enabled" -}}
+{{- $noproxy = append $noproxy (print ((index .Values "kubeclarity-trivy-server" "service" "name") | default (include "trivy.fullname" (index .Subcharts "kubeclarity-trivy-server"))) "." .Release.Namespace ":" (index .Values "kubeclarity-trivy-server" "service" "port")) -}}
+{{- end -}}
```
sambetts-cisco, line 108: Generally its good practise to define variables at the top of the template just to prevent them affecting indentation and also to keep the template as readable as possible. There are a few other templates in the chart where you can find this being done. Would you mind moving this up please.

sambetts-cisco (changes requested):
One tiny change request to move the variable definition to the top of the file, otherwise LGTM! Once that has been fixed I'll be happy to approve

charts/kubeclarity/templates/scanner-template-configmap.yaml
florentinadolf: done

sambetts-cisco (approved):
Thanks @florentinadolf! LGTM!

PR #309, opened 2023-01-20
Prometheus metrics improvements

nostra:
2 things changed with this PR:

1: It turns out that I cannot use label "severity" as it clashes with the following, making alerts complain:
https://github.com/prometheus-operator/kube-prometheus/blob/0ac0d7f9bf5ab1de3e596ff3bdd4b36c52e27267/jsonnet/kube-prometheus/components/alertmanager.libsonnet#L52
This makes alert complain: `"vector contains metrics with the same labelset after applying alert labels"`

Renaming the label `severity` to `vul_severity` to avoid this.

2: Added an initial call to `s.recordVulnerabilities` in order to prime the metrics at startup. Otherwise, metrics will make a dip to zero upon server restart.

### DIFF
--- a/backend/pkg/metrics/README.md
+++ b/backend/pkg/metrics/README.md
@@ -16,17 +16,17 @@ kubeclarity:
 
 The metrics that are exposed are:
 
-| Metric                                               | Explanation                                                        |
-|------------------------------------------------------|--------------------------------------------------------------------|
-| kubeclarity_application_vulnerability                | Count of vulnerabilities per application, environment and severity |
-| kubeclarity_number_of_applications                   | The total number of applications                                   |
-| kubeclarity_number_of_fixable_vulnerabilities        | The number of fixable vulnerabilities per severity                 |
-| kubeclarity_number_of_fixable_vulnerabilities_amount | The total number of fixable vulnerabilities                        |
-| kubeclarity_number_of_packages                       | The total number of packages                                       |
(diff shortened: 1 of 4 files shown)


backend/pkg/metrics/README.md
```diff
@@ -16,17 +16,17 @@ kubeclarity:
...
-| kubeclarity_number_of_vulnerabilities                | The number of vulnerabilities per severity                         |
-| kubeclarity_number_of_vulnerabilities_amount         | The total number of vulnerabilities                                |
-| kubeclarity_vulnerability_trend                      | Vulnerability trend in a 60 minute time window                     |
+| Metric                                               | Explanation                                                                          |
+|------------------------------------------------------|--------------------------------------------------------------------------------------|
+| kubeclarity_application_vulnerability                | Count of vulnerabilities per application, environment and importance (i.e. severity) |
+| kubeclarity_number_of_applications                   | The total number of applications                                                     |
+| kubeclarity_number_of_fixable_vulnerabilities        | The number of fixable vulnerabilities per importance (i.e. severity)                 |
```
sambetts-cisco, line 23: The industry standard for vulnerabilities severity is severity so it would be nice to avoid using a different word to describe it, I personally don't know enough about prometheus to know the answer here, but can we have a label like "vul_severity"? If not "seriousness" might be a better descriptor than importance here I think.

FrimIdan (approved):

PR #313, opened 2023-02-02
Upgrade postgres chart to 11.9.13 and upgrade image to 14.6.0-debian-11-r31

sambetts-cisco:
This upgrades postgres to the latest chart that supports postgres 14 and latast postgres 14 image from bitnami. The new image fixes a large number of vulnerabilities. We can't move to chart 12.X.X as that picks up postgres 15 which is incompatible with postgres 14 during an upgrade.

PR #314, opened 2023-02-02
Update postgres chart to 12.1.14 including postgres 15.1

sambetts-cisco:
This will require manual upgrade of the postgres DB, TBD how we're going
to do that, maybe init containers or something.


FrimIdan:
following https://github.com/openclarity/kubeclarity/pull/313#pullrequestreview-1281182890 not sure we need to upgrade to postgres 15.1

sambetts-cisco:
> following [#313 (review)](https://github.com/openclarity/kubeclarity/pull/313#pullrequestreview-1281182890) not sure we need to upgrade to postgres 15.1

Yeah I agree, the only thing we might need to be careful about is that it doesn't appear that bitnami are supporting the 11.X.X chart any longer, but they are publishing images for postgres 14.X

PR #315, opened 2023-02-03
Resolve ImagePullSecrets in the ScannerJob not in backend

### LINKED ISSUE #296: Remove cluster scope secrets read permissions
Opened following https://github.com/openclarity/kubeclarity/pull/287

In the current implementation, KubeClarity SA needs cluster scope secrets read permissions for the cases where a pod have multiple imagePullSecrets and we need to find the matching one ([here](https://github.com/openclarity/kubeclarity/blob/main/runtime_scan/pkg/scanner/scanner.go#L202)  and [here](https://github.com/openclarity/kubeclarity/blob/main/runtime_scan/pkg/scanner/scanner.go#L231)) 

We need to look if that logic can be skipped and try to mount all imagePullSecrets into the scanner job and the scanner job will handle multiple secrets in order to find the right creds.

Need to see that the mount is **optional** so a scanner job will not be blocked from running if a secret is missing (imagePullSecrets can hold secret names that are not exists)

If that approach works, we need to have only Role for read secrets for the CREDS_SECRET_NAMESPACE (to support credentials put by users to the pre-define secrets in CREDS_SECRET_NAMESPACE namespace)

Also, while checking the above, we can see if other permissions can move to be namespace scope and not cluster scope (ConfigMap for example)

Once done, we should
1. Update the `readClusterSecrets` to be namespace scoped related
2. Update the `KubeClarity requires these K8s permissions:` Note in the README

sambetts-cisco:
To remove the requirement for KubeClarity to read secrets from the whole cluster, mount all the ImagePullSecrets available to the target pod, and then resolve the correct authentication credentials in the scanner job.

Fixes #296


sambetts-cisco:
@MasayaAoyama please take a look at this PR, I'm dropping the environment variable and code you added to prevent secret reads from the cluster, however going forward the KubeClarity service account and backend will only be able to read secrets from the CREDS_SECRET_NAMESPACE hopefully that satisfies your requirements still?

charts/kubeclarity/templates/role.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "kubeclarity.name" . }}
```
FrimIdan, line 4: missing `namespace: '{{ .Release.Namespace }}'`

charts/kubeclarity/templates/rolebinding.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ include "kubeclarity.name" . }}
```
FrimIdan, line 4: missing `namespace: '{{ .Release.Namespace }}'`

runtime_scan/pkg/scanner/job_managment.go
```diff
@@ -328,30 +330,22 @@ func removeCISDockerBenchmarkScannerFromJob(job *batchv1.Job) {
 // 2. Mount the volume into each container to a specific path (`BasicVolumeMountPath`/`DockerConfigFileName`)
 // 3. Set `DOCKER_CONFIG` to point to the directory that contains the config.json.
 func setJobDockerConfigFromImagePullSecret(job *batchv1.Job, secretName string) {
```
FrimIdan, line 332: function comment should be updated?

runtime_scan/pkg/scanner/job_managment.go
```diff
@@ -328,30 +330,22 @@ func removeCISDockerBenchmarkScannerFromJob(job *batchv1.Job) {
...
 		container.VolumeMounts = append(container.VolumeMounts, corev1.VolumeMount{
-			Name:      _creds.BasicVolumeName,
+			Name:      volumeName,
 			ReadOnly:  true,
-			MountPath: _creds.BasicVolumeMountPath,
-		})
-		container.Env = append(container.Env, corev1.EnvVar{
-			Name:  _creds.DockerConfigEnvVar,
-			Value: _creds.BasicVolumeMountPath,
+			MountPath: fmt.Sprintf("/opt/kubeclarity-pull-secrets/%s", secretName),
```
FrimIdan, line 348: `/opt/kubeclarity-pull-secrets`
should be const

runtime_scan/pkg/scanner/scanner.go
```diff
@@ -202,8 +199,13 @@ func (s *Scanner) initScan() error {
 			continue
 		}
 
-		// TODO: (idanf) verify if we need to read pod image pull secrets or just mount it to the scanner job
-		secrets := k8sutils.GetPodImagePullSecrets(s.clientset, pod)
+		// Fetch all image pull secrets from the pod so that we can
+		// inform the scanner where to check for credentials to pull
+		// the image to scan.
+		imagePullSecretNames := []string{}
+		for _, ips := range pod.Spec.ImagePullSecrets {
```
FrimIdan, line 206: now that we are not trying to fetch the secrets, we don't know if secret is missing, we need to do the mount secret `optional`. we don't want that the scanner will failed to run because of that (and we do see from time to time that secret in image pull secret is missing)

runtime_scan/pkg/scanner/scanner.go
```diff
@@ -229,13 +231,13 @@ func (s *Scanner) initScan() error {
...
-				podName:         pod.GetName(),
-				namespace:       pod.GetNamespace(),
-				imagePullSecret: k8sutils.GetMatchingSecretName(secrets, container.Image),
-				imageName:       container.Image,
-				podUID:          string(pod.GetUID()),
-				podLabels:       labels.Set(pod.GetLabels()),
+				containerName:    container.Name,
+				podName:          pod.GetName(),
+				namespace:        pod.GetNamespace(),
+				imagePullSecrets: imagePullSecretNames,
```
FrimIdan, line 237: please delete `GetMatchingSecretName` and `GetPodImagePullSecrets` functions if not used now.

shared/pkg/config/runtime_scanner.go
```diff
@@ -40,19 +49,131 @@ func setRegistryConfigDefaults() {
...
-	username, password, err := creds.ExtractCredentials(imageID)
-	if err != nil {
-		log.Warnf("Failed to extract credentials for imageID %v: %v", imageID, err)
+	var auth Auth
+
+	imagePullSecretPath := viper.GetString(ImagePullSecretPath)
+	if imagePullSecretPath != "" {
+		authConfig, err := GetAuthConfigFromPullSecrets(imagePullSecretPath, imageID)
+		if err != nil {
+			log.Fatalf("Failed to resolve auth config from pull secrets: %v", err)
```
FrimIdan, line 58: Before we just printed a warning and continued, i think we should do the same here. At worst we will fail to scan due to missing credentials, at best we will succeed :)

shared/pkg/config/runtime_scanner.go
```diff
@@ -40,19 +49,131 @@ func setRegistryConfigDefaults() {
...
 		UseHTTP:       viper.GetBool(RegistryUseHTTP),
-		Auths: Auths{
-			{
-				Username: username,
-				Password: password,
+		Auths:         []Auth{auth},
+	}
+}
+
+func GetAuthConfigFromPullSecrets(ipsPath string, imageName string) (*authn.AuthConfig, error) {
```
FrimIdan, line 81: seems like `GetAuthConfigFromPullSecrets` and `NewKeyChainFromPath` should be in a util pkg, DYT?

shared/pkg/config/runtime_scanner.go
```diff
@@ -40,19 +49,131 @@ func setRegistryConfigDefaults() {
 func LoadRuntimeScannerRegistryConfig(imageID string) *Registry {
 	setRegistryConfigDefaults()
 
-	username, password, err := creds.ExtractCredentials(imageID)
-	if err != nil {
-		log.Warnf("Failed to extract credentials for imageID %v: %v", imageID, err)
+	var auth Auth
```
FrimIdan, line 52: so we are not using `ExtractCredentials` any more? we are not using the `DefaultKeychain` that uses the docker config file?
Can't we use a single docker config file with multiple creds in it?

sambetts-cisco: The job doesn't have a docker config file outside of the ones mounted in by the secret, so I dropped ExtractCredentials and instead copied the logic that kubelet/go-containerregistry does here: https://github.com/google/go-containerregistry/blob/main/pkg/authn/kubernetes/keychain.go#L63-L134 except instead of loading the secrets from kubernetes I'm loading them from disk.

shared/pkg/config/runtime_scanner.go
sambetts-cisco: moving these into the existing shared/utils/creds pkg 👍

shared/pkg/config/runtime_scanner.go
FrimIdan: ok, make sense. so we should delete `ExtractCredentials` and all code related.

sambetts-cisco: I added a commit which removes the dead code including ExtractCredentials 👍

runtime_scan/pkg/scanner/job_managment.go
```diff
@@ -322,40 +325,51 @@ func removeCISDockerBenchmarkScannerFromJob(job *batchv1.Job) {
...
-			Name:  _creds.DockerConfigEnvVar,
-			Value: _creds.BasicVolumeMountPath,
+			MountPath: path.Join(imagePullSecretMountPath, secretName),
 		})
 	}
 }
 
+// Set the IMAGE_PULL_SECRET_PATH environment variable to
+// /opt/kubeclarity-pull-secrets.
+func setJobImagePullSecretPath(job *batchv1.Job) {
```
akpsgit, line 366: maybe better to check if it doesn't exist first

shared/pkg/config/runtime_scanner.go
```diff
@@ -40,19 +41,32 @@ func setRegistryConfigDefaults() {
...
+
+	imagePullSecretPath := viper.GetString(ImagePullSecretPath)
+	if imagePullSecretPath != "" {
+		authConfig, err := creds.GetAuthConfigFromPullSecretPath(imagePullSecretPath, imageID)
+		if err != nil {
+			// Just log the issue don't fail, getting the credentials is best effort
+			log.Warnf("Failed to resolve auth config from pull secrets: %v", err)
```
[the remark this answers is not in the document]
sambetts-cisco: good catch, thanks!

MasayaAoyama:
@sambetts-cisco 

sorry for late.

> however going forward the KubeClarity service account and backend will only be able to read secrets from the CREDS_SECRET_NAMESPACE hopefully that satisfies your requirements still?

I saw your PR. I think my requirements is still satisfied.
I want to scan all images in the cluster using cloud credential(e.g. GCP service account json key) which is in a "kubeclarity" namespace.

sambetts-cisco:
@MasayaAoyama awesome, glad we're not introducing a regression on your original PR, thanks for looking!

akpsgit (dismissed):

runtime_scan/pkg/scanner/creds/basic.go
```diff
@@ -54,37 +58,34 @@ func (u *BasicRegCred) ShouldAdd() bool {
...
-		})
-		container.Env = append(container.Env, corev1.EnvVar{
-			Name:  DockerConfigEnvVar,
-			Value: BasicVolumeMountPath,
+			MountPath: fmt.Sprintf("/opt/kubeclarity-pull-secrets/%s", BasicRegCredSecretName),
 		})
 	}
+
+	// Configure the IMAGE_PULL_SECRET_PATH to the location where we mounted the image pull secret
+	for i := range job.Spec.Template.Spec.Containers {
```
fishkerez, line 87: Why do we need another for loop here and doing this in the previous one?

sambetts-cisco: Great catch, I had copied this from the job_management logic but there is no need for a double loop

pbalogh-sa (approved):

Issue #353, opened 2023-02-21
Feature request: configure max scan parallelism


MasayaAoyama:
**Is your feature request related to a problem? Please describe.**

large scanner job is running at a once, so cluster is auto scaled to large number in a moment.
After job will be completed, cluster have a too many free resources until cluster is down scaled.

**Describe the solution you'd like**

We would like to configure parrallism for running scanner job.
But currently, `MaxScanParallelism` is set to 10 as magic number.

https://github.com/openclarity/kubeclarity/blob/01a8c4a1522967acd93341d32274ab8c820279af/backend/pkg/runtime_scanner/scanner.go#L180

I think we can two options as following.

**Describe alternatives you've considered**

I would like to following,

1. set default value with cli options.
2. set specific number via Web UI (include periodically execution). 

What do you think of this proposal?
I think it can be implement both.

**Additional context**

I implemented sample code for pattern 1.

FrimIdan:
Done in https://github.com/openclarity/kubeclarity/pull/355

PR #355, opened 2023-02-21
implement handling default parallelism for scan job

MasayaAoyama:
sample implementations for https://github.com/openclarity/kubeclarity/issues/353
It can be treat max scan parallelism from manifests(env var) with viper.


backend/pkg/config/config.go
```diff
@@ -38,13 +38,16 @@ const (
 	EnableDBInfoLogs          = "ENABLE_DB_INFO_LOGS"
 	ViewRefreshIntervalEnvVar = "DB_VIEW_REFRESH_INTERVAL"
 
+	DefaultScanParallelism = "DEFAULT_SCAN_PARALLELISM"
```
FrimIdan, line 41: I think it will be useful to be able to configure it in the helm charts & values, WDYT?

MasayaAoyama: I think so too 👍🏻

MasayaAoyama: @FrimIdan I think so too 👍🏻 I pushed.

backend/pkg/backend/backend.go
```diff
@@ -106,7 +106,7 @@ func Run() {
 		log.Fatalf("Failed to create runtime scan orchestrator: %v", err)
 	}
 
-	restServer, err := rest.CreateRESTServer(config.BackendRestPort, backend.dbHandler, orc, k8sClientset)
+	restServer, err := rest.CreateRESTServer(config.BackendRestPort, backend.dbHandler, orc, k8sClientset, config.DefaultScanParallelism)
```
FrimIdan, line 109: IMO it will be better to set the default during `CreateScanner` (called from `createRuntimeScanOrchestrator`-->`orchestrator.Create`) and set it on `Scanner`. 
So the default value should be set on runtime_scan/pkg/config and not on the general config.
When we get into `(s *Scanner) jobBatchManagement()`  
```
	numberOfWorkers := s.scanConfig.MaxScanParallelism
```
`numberOfWorkers` - should be set from `scanConfig` and if empty should be taken from `s.defaultScanParallelism`

It will remove all the propegation of the default value in less relevant points like `rest.CreateRESTServer`.

WDYT? Hope the comment was clear enough.

MasayaAoyama: @FrimIdan 
Thank you for your comment :)
I implemented again.

backend/pkg/runtime_scanner/scanner.go
```diff
@@ -176,8 +179,16 @@ func (s *RuntimeScanner) startScan(scanConfig *ScanConfig) error {
...
 	}
 
+	var parallelism int
+	// If set specific parallelism from web UI, use that
+	if scanConfig.Parallelism != 0 {
+		parallelism = scanConfig.Parallelism
+	} else {
+		parallelism = s.defaultScanParallelism
+	}
+
```
FrimIdan, line 189: `defaultScanParallelism` should be removed from `RuntimeScanner` and `scanConfig.Parallelism` should be set directly to `runtime_scan_config.ScanConfig.MaxScanParallelism`.
I don't think we even need the `Parallelism` on `scanConfig` since we don't have an API support, but if you decided to keep it, please put a TODO to handle it from the UI.

backend/pkg/runtime_scanner/scanner.go
```diff
@@ -54,6 +54,8 @@ type RuntimeScanner struct {
 	// New scan requests are coming through this channel.
 	scanChan chan *ScanConfig
 	lock     sync.RWMutex
+
+	defaultScanParallelism int
```
FrimIdan, line 58: remove

backend/pkg/backend/backend.go
FrimIdan: @MasayaAoyama Thank you, it looks a lot better now, just added a small comment and we good to go I think.

MasayaAoyama:
@FrimIdan 
sorry for changing implementation.
I added default parallelism values to `runtime_scan`, but I think if we should to configure parallelism via API/Web UI, we cannot handling default parallelism value.
because, `s.scanConfig.MaxScanParallelism` is always configure from DB, so default parallelism values are ignored.
So I re-implemented to add API and webui.

If you want to add default parallelism parameter instead of this implementation (modify API and web UI), I can revert this.

# Notes

* models.RuntimeQuickScanConfig
  * save following to DB
    * CisDockerBenchmarkScanEnabled
    * MaxScanParallelism
  * (not save to DB, not have field at models.RuntimeQuickScanConfig) configure at posting data to server
    * Namespaces
* models.RuntimeScheduleScanConfig
  * save following to DB
    * CisDockerBenchmarkScanEnabled
    * MaxScanParallelism
    * Namespaces

# Screenshots

image

image

FrimIdan:
@MasayaAoyama Thanks for the update, it really looks better now, LGTM.
@galiail can you please take a look at the small ui changes?

MasayaAoyama:
@FrimIdan 
E2E test was passed at my local environment, and I think e2e test is correct maybe.
So can I retest e2e actions?
I cannot re-run github actions.

FrimIdan:
> hmm... e2e test was not passed... Do you have suspicious point or any idea?
> 
> On my local linux machine, e2e test was passed with following steps.
> 
> * delete `e2e/kubeclarity-cli` and `cli/bin/cli`
> * delete all docker images
> * delete go test cache `go clean -testcache`
> * build cli and docker images `make cli` `make docker`
> * run `make e2e`

The steps to run e2e described [here](https://github.com/openclarity/kubeclarity/blob/main/CONTRIBUTING.md#testing-end-to-end) - did you updated the `charts/kubeclarity/values.yaml` with the correct tag?
BTW - you probably want to update `MaxScanParallelism` in the following e2e function, the default is 1, right?
```
func startRuntimeScan(t *testing.T) {
	t.Helper()
	_ = common.PutRuntimeQuickScanConfig(t, kubeclarityAPI, &models.RuntimeQuickScanConfig{
		CisDockerBenchmarkScanEnabled: true,
	})
	_ = common.PutRuntimeScanStart(t, kubeclarityAPI, &models.RuntimeScanConfig{
		Namespaces: []string{"test"},
	})
}
```

MasayaAoyama:
@FrimIdan 
Thank you for your review.
I have thought `MaxScanParallelism` is set to 10 at initialization, but it was incorrect.
I set parallelism explicitly, e2e was passed.

FrimIdan (approved):

MasayaAoyama:
@FrimIdan
Hi, do you have any plan to release new version include this PR?

sambetts-cisco:
Hey @MasayaAoyama, I've just tagged v2.15.1 which should include your change plus a couple of other fixes which merged, sorry for the delay!

PR #360, opened 2023-02-28
do not normalize local images

fishkerez:


shared/pkg/utils/image/image.go
```diff
@@ -53,3 +59,11 @@ func imageNameWithDigestOrTag(named reference.Named) reference.Named {
 
 	return named
 }
+
+func ValidateLocalImageID(imageID string) error {
```
FrimIdan, line 63: IMO it's better that the function will return a boolean and each called will handle it instead of ignoring the returned error in some cases

FrimIdan (approved):

FrimIdan (approved):

PR #387, opened 2023-04-02
Gzip SBOM

FrimIdan:
## Description

On huge sbom (~10M) scans, the sbom-db got OOMKilled.
```
  Warning  OOMKilling                              10m                kernel-monitor                     Memory cgroup out of memory: Killed process 13698 (sbom_db) total-vm:808280kB, anon-rss:101024kB, file-rss:11708kB, shmem-rss:0kB, UID:1000 pgtables:316kB oom_score_adj:998
```

This PR solves the issue with 2 fixes:
1. Update the sbom-db resources limits
2. Save a gzipped base64 string of the sbom, reducing the amount of bytes going thorough the wire and being marshaled/unmarshaled  

## Type of Change

[X] Bug Fix
[ ] New Feature
[ ] Breaking Change
[X] Refactor
[ ] Documentation
[ ] Other (please describe)

## Checklist


charts/kubeclarity/values.yaml
```diff
@@ -316,8 +316,8 @@ kubeclarity-sbom-db:
       memory: "20Mi"
       cpu: "10m"
     limits:
-      memory: "100Mi"
-      cpu: "100m"
+      memory: "1Gi"
```
akpsgit, line 319: Were the memory spikes disappeared after gzip? If so, why do we need this?

shared/pkg/utils/gzip/gzip.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"bytes"
+	"compress/gzip"
+	"encoding/base64"
+	"fmt"
+	"io"
+)
+
+// Compress gzip and base64 encode the source input.
```
akpsgit, line 26: Compress (gzip) and encode (base64) the source input

shared/pkg/utils/gzip/gzip.go
```diff
@@ -0,0 +1,65 @@
...
+func Compress(source []byte) (string, error) {
```
akpsgit, line 27: maybe better CompressAndEncode

shared/pkg/utils/gzip/gzip.go
```diff
@@ -0,0 +1,65 @@
...
+		return "", fmt.Errorf("failed to gzip flush: %w", err)
+	}
+	if err := gz.Close(); err != nil {
+		return "", fmt.Errorf("failed to gzip close: %w", err)
+	}
+	return base64.StdEncoding.EncodeToString(b.Bytes()), nil
+}
+
+// UnCompress base64 decode and unzip the source input.
+func UnCompress(source string) ([]byte, error) {
```
akpsgit, line 46: maybe DecodeAndUncompress

runtime_k8s_scanner/pkg/sbomdb/getter.go
```diff
@@ -55,5 +56,10 @@ func (g *GetterImpl) Get(ctx context.Context, imageHash string) ([]byte, error)
 		}
 	}
 
-	return sbom.Payload.Sbom, nil
+	uncompressedSbom, err := gzip.UnCompress(sbom.Payload.Sbom)
```
akpsgit, line 59: Maybe we should add a arg to the function whether unzip is needed. We will need a flow where we need to handle it zipped e2e

charts/kubeclarity/values.yaml
FrimIdan: To be on the safe side, low memory limits could result with a kernel oom killer for nothing in case of spikes of memory that we just had. It doesn't required more from the nodes, it just to prevent memory leakage over time.

runtime_k8s_scanner/pkg/sbomdb/getter.go
FrimIdan: > We will need a flow where we need to handle it zipped e2e

Not sure in what cases, I wanted to make the Get/Set functions unaware of the gzip/unzip logic, also to not break the existing interface.

akpsgit: We can add a new GetCompressed method

sbom_db/api/swagger.yaml
```diff
@@ -33,7 +33,7 @@ definitions:
     properties:
       sbom:
         type: 'string'
-        format: 'byte'
+        description: 'base64 gzipped sbom'
```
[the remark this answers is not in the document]
FrimIdan: yes, we wanted to avoid the json encoce/decode to the jobs since we don't need to store it decoded in the DB, it's a waste of cycles and memory.

Issue #389, opened 2023-04-06
Execute scanner jobs in kubeclarity's namespace rather than in originating pod's


giovannirco:
**Is your feature request related to a problem? Please describe.**
Each scanner job will be created on originating pod namespace, fetching logs as an example makes it hard as I need to specify each namespace rather than just matching a label

**Describe the solution you'd like**
A toggle on helm values to specify which namespace to spawn jobs and a blank value to leave it as it is now for backwards compatibility

**Describe alternatives you've considered**
I was thinking about using kyverno policies to mutate the namespace where the scanners will run 

**Additional context**

I have cilium installed in my cluster and I need to create rules for each namespace to allow scanners to do its job
I was also playing around with IRSA to allow scanners to fetch ECR images in a more secure manner
For both tasks having the scanners jobs in kubeclarity namespace would make things much easier

FrimIdan:
Hi @giovannirco thanks for opening the issue.
The idea behind running the scanner job on the pod namespace is to be able to support scanning images that required image pull secrets to fetch the image, since the secret is in the same namespace of the pod.
Having said that, we can add a support to set the default scanners jobs namespace in case image pull secrets are not used.

PR #416, opened 2023-05-10
Add a support to set the default scanners namepace

FrimIdan:
## Description

The PR adds a support to be able to set the default namespace where job scanners should run.
If a default value is set it may be overridden if:
1. The scanner job must run in the pod namespace to fetch image pull secrets, OR
2. The scanner job must run in the release namespace to fetch service credentials (more info in https://github.com/openclarity/kubeclarity#private-registries-support-for-k8s-runtime-scan)

Solves https://github.com/openclarity/kubeclarity/issues/389

## Type of Change

[ ] Bug Fix
[X] New Feature
[ ] Breaking Change
[ ] Refactor
[ ] Documentation
[ ] Other (please describe)

## Checklist

### DIFF
--- a/charts/kubeclarity/values.yaml
+++ b/charts/kubeclarity/values.yaml
@@ -118,6 +118,12 @@ kubeclarity-runtime-scan:
     app: kubeclarity-scanner
     sidecar.istio.io/inject: "false"
 
+  ## Scanner jobs namespace.
+  # If set, the scanner jobs will run in the given namespace unless:
+  # 1. The scanner job must run in the pod namespace to fetch image pull secrets, OR
+  # 2. The scanner job must run in the release namespace to fetch service credentials (more info in https://github.com/openclarity/kubeclarity#private-registries-support-for-k8s-runtime-scan)
+  namespace: ""
+
   ## Scanner pods tolerations.
   # tolerations:
   # - key: key1
--- a/charts/kubeclarity/values.yaml
+++ b/charts/kubeclarity/values.yaml
@@ -119,6 +119,7 @@ kubeclarity-runtime-scan:
     sidecar.istio.io/inject: "false"
 
   ## Scanner jobs namespace.
+  # If left blank, the scanner jobs will run in the same namespace as pod being scanned.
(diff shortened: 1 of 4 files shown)


charts/kubeclarity/values.yaml
```diff
@@ -118,6 +118,12 @@ kubeclarity-runtime-scan:
     app: kubeclarity-scanner
     sidecar.istio.io/inject: "false"
 
+  ## Scanner jobs namespace.
+  # If set, the scanner jobs will run in the given namespace unless:
+  # 1. The scanner job must run in the pod namespace to fetch image pull secrets, OR
+  # 2. The scanner job must run in the release namespace to fetch service credentials (more info in https://github.com/openclarity/kubeclarity#private-registries-support-for-k8s-runtime-scan)
```
sambetts-cisco, line 124: We should probably document that "If left blank, the scanner jobs will run in the same namespace as pod being scanned"

runtime_scan/pkg/scanner/job_managment.go
```diff
@@ -284,19 +284,25 @@ func (s *Scanner) createJob(data *scanData) (*batchv1.Job, error) {
 		removeCISDockerBenchmarkScannerFromJob(job)
 	}
 	job.SetName(jobName)
-	job.SetNamespace(podContext.namespace)
+	if job.Namespace == "" {
+		// Set job namespace to pod namespace if not set in the template.
+		// May be overridden later if support for image pull secrets or SA credentials is needed.
+		job.SetNamespace(podContext.namespace)
```
sambetts-cisco, line 290: @FrimIdan  do you know a reason why we couldn't default to the release namespace here? and then only use the pod namespace for the imagePullSecret case?

FrimIdan: Just because we want to keep the old behavior.

sambetts-cisco: 👍 sounds good

sambetts-cisco (approved):

Issue #413, opened 2023-05-04
Enable the possibility to delete application resources


fluflo:
**Is your feature request related to a problem? Please describe.**
I'm currently building an extension on base of kubeclarity to automatically update vulnerable resources to a newer version. So after I updated the container images to a newer version, the kubeclarity still tracks the older application resource which isn't used by an application. As a result the application resource overview keeps growing and it's getting more difficult to track the versions currently installed on the kubernetes cluster.

**Describe the solution you'd like**
There should be an endpoint, which allows to delete application resources. There might be a rule checking if the application resource is still used by an application, which would result in a 400 bad request error. 

**Describe alternatives you've considered**
There could also be the possibility, that after a scan, the unused application resources are cleaned up.

**Additional context**

![image](https://user-images.githubusercontent.com/29951492/236194589-27de2ecd-25bd-4dcd-b18c-6b2c5987f25e.png)

akpsgit:
Hey @fluflo, thanks a lot for the proposal. Indeed a supper useful feature. We can allow the users whether upon an application resource deletion they want to delete its related objects (e.g. un-associate from all applications and delete its related packages and their related vulnerabilities):
![image](https://user-images.githubusercontent.com/39153880/236666315-bf18b8f6-535a-481b-992e-6d5acb45e7ac.png)

Additionally, it possible for your extension to apply a filter "Applications greater than or equal to 1" to disregard the application resources that are detached from an application. 
Does the issue affect any other usage of kubeclarity other than your extension (e.g. Dashboard, etc.)?

akpsgit:
@fluflo thanks for the clarification. We'll prioritize this to start working on it probably next week.

pbalogh-sa:
@fluflo you can find an example of how you can delete an application resource:
https://github.com/openclarity/kubeclarity/pull/434#issuecomment-1566120839

The UI will be updated soon in order to add a delete button for application resources.

PR #434, opened 2023-05-26
Delete application resources

pbalogh-sa:
## Description

Capability to delete application resources.
- Deletes related objects as well that have no association with other resources.

closes: #413

## Type of Change

[ ] Bug Fix
[x] New Feature
[ ] Breaking Change
[ ] Refactor
[ ] Documentation
[ ] Other (please describe)

## Checklist


backend/pkg/database/join_tables.go
```diff
@@ -139,6 +144,12 @@ func (j *JoinTablesHandler) DeleteRelationships(params DeleteRelationshipsParams
 				return fmt.Errorf("failed to delete relationships in %q join table: %v",
 					ResourcePackagesJoinTableName, err)
 			}
+			ar := FilterIs(tx.Model(&ApplicationResources{}),
+				FieldInTable(ApplicationResourcesJoinTableName, columnJoinTableResourceID), params.ResourceIDsToRemove)
+			if err := ar.Delete(&ApplicationResources{}).Error; err != nil {
+				return fmt.Errorf("failed to delete relationships in %q join table: %v",
+					ApplicationResourcesJoinTableName, err)
+			}
```
FrimIdan, line 152: why it was added now?

backend/pkg/database/join_tables.go
```diff
@@ -166,9 +177,8 @@ func (j *JoinTablesHandler) GetResourcePackageIDToAnalyzers(resourceIDs []string
 	if len(resourceIDs) == 0 {
 		return nil, nil
 	}
-
-	var resourcePackages []ResourcePackages
 	resourcePkgIDToAnalyzers := make(map[ResourcePkgID][]string)
+	var resourcePackages []ResourcePackages
```
FrimIdan, line 181: why it was moved?

backend/pkg/database/package.go
```diff
@@ -121,6 +123,29 @@ func (p *PackageTableHandler) Create(pkg *Package) error {
...
+	if err := p.packagesTable.Delete(pkg).Error; err != nil {
+		return fmt.Errorf("failed to delete package: %v", err)
+	}
+
+	p.viewRefreshHandler.TableChanged(packageTableName)
+
+	return nil
+}
+
+func (p *PackageTableHandler) GetDBPackage(id string) (*Package, error) {
```
FrimIdan, line 136: similar to `GetDBResource(id string, shouldGetRelationships bool)` we don't want/need to get the package relationships?

backend/pkg/rest/applications_resources_controller.go
```diff
@@ -71,3 +73,162 @@ func (s *Server) GetApplicationResources(params operations.GetApplicationResourc
...
+	if err != nil {
+		log.Errorf("application resource not found by id %q: %v", id, err)
+		return operations.NewDeleteApplicationResourcesIDNotFound()
+	}
+
+	// Get package IDs that has relations with the resource.
+	relatedPackageIDs, err := s.getRelatedPackageIDs(resource.ID)
+	if err != nil {
+		log.Errorf("Failed to get packages for resource ID=%s: %v", resource.ID, err)
+		operations.NewDeleteApplicationResourcesIDNotFound()
```
FrimIdan, line 89: `operations.NewDeleteApplicationResourcesIDNotFound()`?
Not sure that is the return we need, also missing a `return` call, no?

backend/pkg/rest/applications_resources_controller.go
```diff
@@ -71,3 +73,162 @@ func (s *Server) GetApplicationResources(params operations.GetApplicationResourc
...
+		log.Error(fmt.Sprintf("Failed to delete application relationships: %v", err))
+		return operations.NewDeleteApplicationResourcesIDDefault(http.StatusInternalServerError).
+			WithPayload(oopsResponse)
+	}
+
+	// Get list of package IDs that has no any relations to other resources.
+	packageIDsToDelete, err := s.getPackagesToDelete(relatedPackageIDs)
+	if err != nil {
+		log.Errorf("Failed to get remaining packages: %v", err)
+		operations.NewDeleteApplicationResourcesIDNotFound()
```
FrimIdan, line 105: same as before

backend/pkg/rest/applications_resources_controller.go
```diff
@@ -71,3 +73,162 @@ func (s *Server) GetApplicationResources(params operations.GetApplicationResourc
...
+		log.Errorf("Failed to get vulnerabilities for packages: %v", err)
+		return operations.NewDeleteApplicationResourcesIDDefault(http.StatusInternalServerError).
+			WithPayload(oopsResponse)
+	}
+
+	// Delete only package<->vulnerabilities relationships.
+	if err = s.dbHandler.JoinTables().DeleteRelationships(database.DeleteRelationshipsParams{
+		PackageIDsToRemove: packageIDsToDelete,
+	}); err != nil {
+		log.Errorf("Failed to delete application relationships: %v", err)
```
FrimIdan, line 120: please fix error print (`application`)

backend/pkg/rest/applications_resources_controller.go
```diff
@@ -71,3 +73,162 @@ func (s *Server) GetApplicationResources(params operations.GetApplicationResourc
...
+
+	return vulnerabilityIDsToDelete, nil
+}
+
+func (s *Server) deletePackages(packageIDs []string) {
+	for _, pkgID := range packageIDs {
+		pkg, err := s.dbHandler.PackageTable().GetDBPackage(pkgID)
+		if err == nil {
+			if err = s.dbHandler.PackageTable().Delete(pkg); err != nil {
+				log.Errorf("Failed to delete package ID=%s: %v", pkgID, err)
```
FrimIdan, line 215: not sure that we need to get before we delete, if the delete function will just get an ID, i'm pretty sure we can just delete by id.
https://gorm.io/docs/delete.html#Delete-with-primary-key
We can create a DeleteByIDs, and delete in a batch.
```
db.Delete(&users, []int{1,2,3})
```

backend/pkg/rest/applications_resources_controller.go
```diff
@@ -71,3 +73,162 @@ func (s *Server) GetApplicationResources(params operations.GetApplicationResourc
...
+	}
+}
+
+func (s *Server) deleteVulnerabilities(vulnerabilityIDs []string) {
+	for _, vulID := range vulnerabilityIDs {
+		vul, err := s.dbHandler.VulnerabilityTable().GetDBVulnerability(vulID)
+		if err == nil {
+			if err = s.dbHandler.VulnerabilityTable().Delete(vul); err != nil {
+				log.Errorf("Failed to delete vulnerability ID=%s: %v", vulID, err)
+			}
```
FrimIdan, line 229: same

backend/pkg/rest/server.go
```diff
@@ -116,6 +116,8 @@ func CreateRESTServer(port int, dbHandler *database.Handler, scanner orchestrato
 		return nil, fmt.Errorf("failed to load swagger spec: %v", err)
 	}
 
+	log.Infof("")
```
FrimIdan, line 119: ?

FrimIdan:
@pbalogh-sa do we want to delete all application resources once an application is deleted and there are orphan application resources?

backend/pkg/rest/applications_resources_controller.go
```diff
@@ -71,3 +73,162 @@ func (s *Server) GetApplicationResources(params operations.GetApplicationResourc
...
+	}
+
+	if err = s.dbHandler.ResourceTable().Delete(resource); err != nil {
+		log.Errorf("Failed to delete application from DB: %v", err)
+		return operations.NewDeleteApplicationResourcesIDDefault(http.StatusInternalServerError).
+			WithPayload(oopsResponse)
+	}
+
+	s.deletePackages(packageIDsToDelete)
+	s.deleteVulnerabilities(vulnerabilityIDsToDelete)
```
FrimIdan, line 140: maybe we want to delete the related application if not connected to another resource?

backend/pkg/database/join_tables.go
pbalogh-sa: removing the resource_packages relationship too not just the applications_resources relations

backend/pkg/database/join_tables.go
pbalogh-sa: I move back

backend/pkg/rest/applications_resources_controller.go
pbalogh-sa: thx

backend/pkg/rest/applications_resources_controller.go
pbalogh-sa: We can use an existing application and attach new scans for it. WDYT @akpsgit

api/swagger.yaml
```diff
@@ -1270,6 +1270,19 @@ paths:
             $ref: '#/definitions/ApplicationResourceEx'
         default:
           $ref: '#/responses/UnknownError'
+    delete:
+      summary: 'Delete Application Resources'
```
akpsgit, line 1274: Delete Application Resource?

backend/pkg/rest/applications_resources_controller.go
akpsgit: I don't think that we should. If the user created it manually and then scanned an image and deleted it, i wouldn't expect that the user need to add the application again. Maybe it's more relevant for automatically created applications, but let's leave it out. WDYT?

backend/pkg/rest/applications_resources_controller.go
```diff
@@ -71,3 +73,150 @@ func (s *Server) GetApplicationResources(params operations.GetApplicationResourc
 			Total: &total,
 		})
 }
+
+func (s *Server) DeleteApplicationResource(id string) middleware.Responder {
```
akpsgit, line 77: are you sure that you want to return in every failure here and leave half dead things behind? maybe it's better to delete as much as we can?

pbalogh-sa: Deleting objects should be in this order because for example if we cannot delete relationships, later we won't be able to delete the objects due to foreign keys in the join tables. Each operation works properly in that case the previous one succeeded.

akpsgit (approved):

PR #443, opened 2023-06-05
Add min value validation on Max Scan Parallelism field

### LINKED ISSUE #441: Max Scan Parallelism parameter under runtime scan config should be a non negative integer
negative values don't make sense for max parallelism parameter
Go to runtime scan config enable the parameter under Options CIS Benchmarks or Scheduled Scan pane from the UI
(e.g. `kubectl logs -n kubeclarity --selector=app=kubeclarity`)

#### Environment:
- Kubernetes version (use `kubectl version --short`):
- Helm version (use `helm version`):
- KubeClarity version (use `kubectl -n kubeclarity exec deploy/kubeclarity -- ./backend version`)
- KubeClarity Helm Chart version (use `helm -n kubeclarity list`)
- Cloud provider or hardware configuration:
- Others:
Screenshot 2023-06-01 at 8 26 03 AM

Screenshot 2023-06-01 at 8 26 32 AM

jashmehta3300:
## Description

Initially the UI allowed negative entries to be entered in the Max Scan Parallelism (under the scan configuration). This is not valid since negative numbers for parallelism  (e.g. Max Scan Parallelism = -20) do not make sense and are not valid according to the API spec. After this change, UI will allow only positive entries to be entered. This fixes #441.

## Type of Change

[x] Bug Fix
[ ] New Feature
[ ] Breaking Change
[ ] Refactor
[ ] Documentation
[ ] Other (please describe)

## Checklist

### DIFF
--- a/ui/src/components/Form/form-fields/TextField/index.js
+++ b/ui/src/components/Form/form-fields/TextField/index.js
@@ -11,7 +11,7 @@ const TextField = ({className, small=false, label, disabled, tooltipText, type="
     return (
         <div className={classnames("form-field-wrapper", "text-field", {small}, className)}>
             {!!label && <FieldLabel tooltipId={`form-tooltip-${field.name}`} tooltipText={tooltipText}>{label}</FieldLabel>}
-            <input type={type} {...field} className="form-field" disabled={disabled} />
+            <input type={type} {...field} className="form-field" disabled={disabled} {...(props.min && { "min": props.min })} />
             {meta.touched && meta.error && <FieldError>{meta.error}</FieldError>}
         </div>
     )
--- a/ui/src/layout/RuntimeScan/ScanConfiguration/ScanOptionsForm.js
+++ b/ui/src/layout/RuntimeScan/ScanConfiguration/ScanOptionsForm.js
@@ -36,7 +36,7 @@ const ScanOptionsForm = ({onClose}) => {
                     })}
                 >
                     <ToggleField name="cisDockerBenchmarkScanEnabled" label="CIS Docker Benchmark" />
-                    <TextField name="maxScanParallelism" label="Max Scan Parallelism" type="number" />
+                    <TextField name="maxScanParallelism" label="Max Scan Parallelism" type="number" min="0" />
                 </FormWrapper>
             }
         </Modal>
--- a/ui/src/layout/RuntimeScan/ScanConfiguration/ScanOptionsForm.js
+++ b/ui/src/layout/RuntimeScan/ScanConfiguration/ScanOptionsForm.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { FETCH_METHODS, useFetch } from 'hooks';
-import FormWrapper, { ToggleField, TextField } from 'components/Form';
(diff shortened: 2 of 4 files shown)


FrimIdan:
LGTM @galiail please take a look and approve if possible.

ui/src/layout/RuntimeScan/ScanConfiguration/ScanOptionsForm.js
```diff
@@ -36,7 +36,7 @@ const ScanOptionsForm = ({onClose}) => {
                     })}
                 >
                     <ToggleField name="cisDockerBenchmarkScanEnabled" label="CIS Docker Benchmark" />
-                    <TextField name="maxScanParallelism" label="Max Scan Parallelism" type="number" />
+                    <TextField name="maxScanParallelism" label="Max Scan Parallelism" type="number" min="0" />
```
akpsgit, line 39: 1?

sambetts-cisco: In the backend what MaxScanParallelism means is "maximum number of concurrent scanners running" so 0 means that nothing would get scanned. Perhaps this is bad wording on our part and this field needs to be renamed "Max Concurrent Scanners" or similar? WDYT @akpsgit ?

sambetts-cisco:
all looking good thanks @jashmehta3300! @galiail please can you take a look at the JS code and make sure it's all good from your perspective

ui/src/layout/RuntimeScan/ScanConfiguration/ScanOptionsForm.js
FrimIdan: The API defined it as min 1, so we should use also 1 in the js code I think
https://github.com/openclarity/kubeclarity/blob/main/api/swagger.yaml#L719

ui/src/components/Form/form-fields/TextField/index.js
```diff
@@ -11,7 +11,7 @@ const TextField = ({className, small=false, label, disabled, tooltipText, type="
     return (
         <div className={classnames("form-field-wrapper", "text-field", {small}, className)}>
             {!!label && <FieldLabel tooltipId={`form-tooltip-${field.name}`} tooltipText={tooltipText}>{label}</FieldLabel>}
-            <input type={type} {...field} className="form-field" disabled={disabled} />
+            <input type={type} {...field} className="form-field" disabled={disabled} {...(props.min && { "min": props.min })} />
```
galiail, line 14: This will block the user from entering negative values using the field arrows but the user will still be able to type a negative number in the field. Instead or in addition, we can utilize the validate function of the formik lib to validate: adding validate={value => {... return "errror message"}} to the relevant TextField.

jashmehta3300: Thank you! @galiail
I have made the relevant changes and pushed it. Here is a screenshot that shows the error message when user types a number less than 0.
Screenshot 2023-06-08 at 6 22 12 PM

galiail (approved):

PR #446, opened 2023-06-15
Fix empty licenses bug: Use license name when there is no ID

akpsgit:
Fix empty package licenses bug. This prefers license ID when exist, and will use the license name if present in case there is no ID

### DIFF
--- a/shared/pkg/utils/cyclonedx_helper/cyclonedx_helper.go
+++ b/shared/pkg/utils/cyclonedx_helper/cyclonedx_helper.go
@@ -102,6 +102,8 @@ func GetComponentLicenses(component cdx.Component) []string {
 		// https://github.com/CycloneDX/specification/blob/1.4/schema/bom-1.4.xsd#L1398
 		// https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/
 		if license.License != nil {
+			// A license must either have one of ID OR Name specified according to the CDX spec, otherwise its invalid CDX:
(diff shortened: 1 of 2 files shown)


FrimIdan (approved):

shared/pkg/utils/cyclonedx_helper/cyclonedx_helper.go
```diff
@@ -102,7 +102,11 @@ func GetComponentLicenses(component cdx.Component) []string {
 		// https://github.com/CycloneDX/specification/blob/1.4/schema/bom-1.4.xsd#L1398
 		// https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/
 		if license.License != nil {
-			licenses = append(licenses, license.License.ID)
+			if license.License.ID != "" {
```
sambetts-cisco, line 105: Maybe worth adding a comment here, license must either have one of ID OR Name specified according to the CDX spec, otherwise its invalid CDX:
https://cyclonedx.org/docs/1.4/json/#tab-pane_components_items_licenses_items_oneOf_i0

sambetts-cisco (changes requested):

shared/pkg/utils/cyclonedx_helper/cyclonedx_helper.go
akpsgit: done

sambetts-cisco (approved):

Issue #440, opened 2023-05-30
Grype shows up under content analyzers column in UI under application resources view


pkalapat:
#### What happened:
Grype shows up under content analyzers

#### What you expected to happen:
Grype is a scanner and not an SBOM/Content analyzer, therefore shouldn't be listed as an analyzer in the UI.

#### How to reproduce it (as minimally and precisely as possible):
1. Enable Trivy as a content analyzer in values.yaml, but don't add trivy as the scanner.
1. Install KubeClarity on EKS under kubeclarity name space
2. Install SockShop in the same cluster under sock-shop name space
3. Start runtime scan on sockshop name space
4. Notice that Grype appears under content analyzer/SBOM Analyzer column in Application Resources page.
5. This seems to be an intermittent issue that is not always observed.
6. We didn't observe this issue When we enabled Trivy as the content analyzer and a scanner.  Perhaps the bug involves adding trivy as the content analyzer but not a scanner. This needs to be investigated

#### Are there any error messages in KubeClarity logs?
(e.g. `kubectl logs -n kubeclarity --selector=app=kubeclarity`)
[backend_pod.log](https://github.com/openclarity/kubeclarity/files/11602782/backend_pod.log)

#### Environment:
- Kubernetes version (use `kubectl version --short`):
- Helm version (use `helm version`):
- KubeClarity version (use `kubectl -n kubeclarity exec deploy/kubeclarity -- ./backend version`)
- KubeClarity Helm Chart version (use `helm -n kubeclarity list`)
- Cloud provider or hardware configuration:
- Others:

pkalapat:
Screenshot of UI showing grype under content analyzers attached
<img width="1685" alt="Screenshot 2023-05-26 at 8 18 12 PM" src="https://github.com/openclarity/kubeclarity/assets/3701244/a26d5c62-56bb-41b4-97ae-edc44bcddd91">

jashmehta3300:
**Environment:**

- Kubernetes Version: 1.27.2
- Helm version: 3.12.0
- KubeClarity Helm Chat Version: 2.18.1
- Cloud provider: AWS

**values.yaml file:** 
[values.txt](https://github.com/openclarity/kubeclarity/files/11604772/values.txt)

fishkerez:
Solved by https://github.com/openclarity/kubeclarity/pull/455

fishkerez:
Solved by https://github.com/openclarity/kubeclarity/pull/455

PR #455, opened 2023-07-04
fix: lock package table on update to prevent dup key violation

fishkerez:
## Description

Following https://github.com/openclarity/kubeclarity/issues/440 we saw the reason is due to failure in handling analyzer results. The error we get is "ERROR: duplicate key value violates unique constraint "packages_pkey"".
This is due to the concurrent nature in which we update the packages in the DB. 
Need to lock the table while updating the packages table in order to avoid this.

Currently gorm does not have support for locking of a table, so I implemented this as a raw query.
Currently only postgress is supported, since sqlite is mostly for test envs, and the raw query is more complicated to do with gorm. 

## Type of Change

[* ] Bug Fix


backend/pkg/database/full_tree.go
```diff
@@ -151,3 +154,16 @@ func (o *ObjectTreeHandler) tableChanged(tables ...string) {
...
+
+func (o *ObjectTreeHandler) LockTable(tx *gorm.DB, tableName string) *gorm.DB {
+	switch o.DriverType {
+	case DBDriverTypeLocal:
+		// not supported currently in sqlite
+		return tx
+	case DBDriverTypePostgres:
+		return tx.Exec(fmt.Sprintf("LOCK TABLE %v IN ACCESS EXCLUSIVE MODE", tableName))
+	default:
+		log.Warnf("DB driver not supported: %v", o.DriverType)
```
akpsgit, line 166: Lock table for DB driver is not supported

akpsgit (approved):

PR #469, opened 2023-08-01
Update stale issue workflow

lelia:
## Description

- [Deprecates](https://github.com/probot/stale/pull/430) `StaleBot` configuration (`.github/stale.yml`) 
- Introduces new workflow using GitHub Actions (`.github/workflows/stale.yml`)
- Ports existing options (days until stale, custom labels) from previous config
- Leverages [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow) configuration hosted on [openclarity/.github](https://github.com/openclarity/.github)
- See also: https://github.com/openclarity/.github/pull/16

## Type of Change

[ ] Bug Fix
[ ] New Feature
[ ] Breaking Change
[x] Refactor
[ ] Documentation
[ ] Other (please describe)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,18 @@
...
+  workflow_dispatch:
+
+jobs:
+  mark-stale:
+    uses: openclarity/.github/.github/workflows/_stale.yml@main
+    permissions:
+      issues: write
+      pull-requests: write
+    with:
+      days-until-stale: 60
```
pkalapat, line 15: This looks great! The 60 days default is quite realistic though I think a 90-day period seems a more generous and reasonable approach to take a quarterly perspective into the stale issues. I'm fine with the current default as long as this number can be altered per project community engagement patterns that emerge over time.

pkalapat (commented):
This looks great! The 60 days default is quite realistic though I think a 90-day period seems a more generous and reasonable approach to take a quarterly perspective into the stale issues. I'm fine with the current default as long as this number can be altered per project community engagement patterns that emerge over time.

.github/workflows/stale.yml
lelia: Appreciate the input! We could certainly adjust to 90 days in the future if 60 still feels too short. With the current configuration, issues will still remain open for 75 days _total_, which feels like a reasonable starting point. 

But yes, these values will be trivial to change in the future once the new pattern is in place 🙌

lelia:
@openclarity/kubeclarity-maintainers now that https://github.com/openclarity/.github/pull/16 has been merged, this is ready for review!

sambetts-cisco (approved):

PR #472, opened 2023-08-03
Changed service port to dynamic

AshutoshAM2002:
## Description

I want to change backend ports and targetport, but these are hardcoded, I am making it dynamic in template/service.yaml.
```yaml
spec:
  ports:
    - name: backend
      port: 8080
      protocol: TCP
      targetPort: 8080
```
when we insert: 8080 in DNS means that your service is listening on port 8080
but port 80 is used when you don't insert any port in DNS. I changed the service port from hard coded to dynamic.

## Type of Change

## Checklist

- [x] All code-style checks pass

### DIFF
--- a/charts/kubeclarity/templates/service.yaml
+++ b/charts/kubeclarity/templates/service.yaml
@@ -12,9 +12,9 @@ spec:
   type: {{ .Values.kubeclarity.service.type }}
   ports:
     - name: backend
-      port: 8080
+      port: {{ .Values.kubeclarity.service.port }}
       protocol: TCP
-      targetPort: 8080
+      targetPort: {{ .Values.kubeclarity.service.targetPort }}
     - name: runtime-scan-results
       port: {{ index .Values "kubeclarity-runtime-scan" "resultServicePort" }}
       protocol: TCP


sambetts-cisco (changes requested):
Can you explain your reasoning a little bit better in the PR description please? Why do you need this change?

It isn't enough to change the targetPort in the service, to make it properly configurable then you need to:
* update the values.yaml to have the port and targetPort defined.
* update the deployment to set the BackendRestPort to service.targetPort, and also to update the "ports" section to use targetPort.
* update the ingress template to use the new ports

For backwards compatibility values.yaml should have 8080 defined as the default port and targetPort for the service otherwise people upgrading will be broken.

FrimIdan:
I agree with @sambetts-cisco the PR is far from being complete.
Wouldn't it be enough to make the `port` configurable? So all other changes needed by @sambetts-cisco comment will not be needed? 
Will it not solve your problem @AshutoshAM2002?

AshutoshAM2002:
Dear @sambetts-cisco and @FrimIdan 
I have tested this changes locally. We just need to change service.yaml to make service port and targetPort dynamic. I am closing this PR and creating new PR with making changes in values.yaml.

PR #473, opened 2023-08-03
Helm: Support affinity for kubeclarity / sbom / grype

hamishforbes:
## Description

The current helm chart doesn't allow configuring either `nodeSelector` or `nodeAffinity` (or `podAffinity` for that matter).

This is a particular problem in heterogeneous clusters where you may have Windows or ARM nodes which kubeclarity can't run on.

Adds an open `affinity` value to kubeclarity / sbom / grype to allow arbitrary node / pod affinity configuration.

e.g. 
```yaml
kubeclarity:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: kubernetes.io/os
            operator: In
            values:
            - linux
          - key: kubernetes.io/arch
            operator: In
            values:
            - amd64
```

n.b. The Trivy and Postgres sub-charts already support affinity configuration.

## Type of Change

- [x] New Feature

## Checklist

- [x] Functionality is documented - in `values.yaml`
- [ ] All code style checks pass - no tests for the helm chart unless I've missed them?

### DIFF
--- a/charts/kubeclarity/templates/deployment.yaml
+++ b/charts/kubeclarity/templates/deployment.yaml
@@ -37,6 +37,10 @@ spec:
       {{- end }}
     spec:
       serviceAccountName: {{ include "kubeclarity.name" . }}
+{{- if .Values.kubeclarity.affinity }}
+      affinity:
+{{ toYaml .Values.kubeclarity.affinity | indent 8 }}
+{{- end }}
       initContainers:
         - name: '{{ include "kubeclarity.name" . }}-wait-for-pg-db'
           image: {{ index .Values "kubeclarity-postgresql" "image" "registry" }}/{{ index .Values "kubeclarity-postgresql" "image" "repository" }}:{{ index .Values "kubeclarity-postgresql" "image" "tag" }}
--- a/charts/kubeclarity/templates/grype_server/deployment.yaml
+++ b/charts/kubeclarity/templates/grype_server/deployment.yaml
@@ -24,6 +24,10 @@ spec:
       securityContext:
         fsGroup: 1000
       {{- end }}
+{{- if (index .Values "kubeclarity-grype-server" "affinity") }}
+      affinity:
+{{ toYaml (index .Values "kubeclarity-grype-server" "affinity") | indent 8 }}
+{{- end }}
       containers:
         - name: grype-server
           image: '{{ index .Values "kubeclarity-grype-server" "docker" "imageRepo" }}/grype-server:{{ index .Values "kubeclarity-grype-server" "docker" "imageTag" }}'
--- a/charts/kubeclarity/templates/sbom_db/deployment.yaml
+++ b/charts/kubeclarity/templates/sbom_db/deployment.yaml
@@ -23,6 +23,10 @@ spec:
       securityContext:
         fsGroup: 1000
       {{- end }}
+{{- if (index .Values "kubeclarity-grype-server" "affinity") }}
+      affinity:
+{{ toYaml (index .Values "kubeclarity-grype-server" "affinity") | indent 8 }}
+{{- end }}
       containers:
         - name: sbom-db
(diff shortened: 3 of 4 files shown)


charts/kubeclarity/templates/deployment.yaml
```diff
@@ -37,6 +37,10 @@ spec:
       {{- end }}
     spec:
       serviceAccountName: {{ include "kubeclarity.name" . }}
+{{- if .Values.kubeclarity.affinity }}
+      affinity:
+{{ toYaml .Values.kubeclarity.affinity | indent 8 }}
+{{- end }}
```
sambetts-cisco, line 43: One minor styling thing here. To avoid weird new lines when the template is rendered its better to use nindent instead of indent. Its also nice to align the if statements with the indent of the part of the template that is being added. For example:

```
  spec:
      serviceAccountName: {{ include "kubeclarity.name" . }}
      {{- if .Values.kubeclarity.affinity }}
      affinity: {{- toYaml .Values.kubeclarity.affinity | nindent 8 }}
      {{- end }}
```

sambetts-cisco (changes requested):
Thanks for the PR @hamishforbes! I actually recently ran into the same thing so this is a very welcome addition.

If possible I think we should also add nodeSelector here with a default that only selects nodes with the linux OS and amd64 architecture. 

e.g. in values.yaml we'll have:
```
  nodeSelector:
    kubernetes.io/os: linux
    kubernetes.io/arch: amd64
```
then we'll have
```
{{- if .Values.kubeclarity.affinity }}
nodeSelector: {{- toYaml .Values.kubeclarity.nodeSelector | nindent 8 }}
{{- end }}
```

this will ensure that the kubeclarity helm chart works out of the box in mixed environments, but ensures that its configurable.

Also in addition to the deployment.yaml for the control plane components, you'll need to add the affinity and nodeSelector to the scanner template:

https://github.com/openclarity/kubeclarity/blob/main/charts/kubeclarity/templates/scanner-template-configmap.yaml 

The scanner template is used to create the scanner jobs at runtime so they will run into the same affinity issues if we don't add it there.

charts/kubeclarity/templates/deployment.yaml
hamishforbes: Doesn't seem to make any difference if you chomp the leading whitespace with `{{-` and use `nindent` vs  `{{` and `indent` when I tested, but have changed to keep consistent with the rest of the chart.

Likewise most of the other conditionals in the template are not aligned with the actual data, so I was just trying to keep consistent there too. Have changed it though

hamishforbes:
> If possible I think we should also add nodeSelector here with a default that only selects nodes with the linux OS and amd64 architecture.
> 
> e.g. in values.yaml we'll have:
> 
> ```
>   nodeSelector:
>     kubernetes.io/os: linux
>     kubernetes.io/arch: amd64
> ```
> 
> then we'll have
> 
> ```
> {{- if .Values.kubeclarity.affinity }}
> nodeSelector: {{- toYaml .Values.kubeclarity.nodeSelector | nindent 8 }}
> {{- end }}
> ```
> 
> this will ensure that the kubeclarity helm chart works out of the box in mixed environments, but ensures that its configurable.

Done.
I've added both `nodeSelector` and `affinity` to the global values so they only need to be set once but can be overridden per-app if needed

> Also in addition to the deployment.yaml for the control plane components, you'll need to add the affinity and nodeSelector to the scanner template:

Skipped right over that as soon as I saw `ConfigMap` 🤦‍♂️

Tehsmash (approved):
Thanks for the update @hamishforbes, LGTM. There are some linter errors for the values.yaml that need to be resolved before we can merge.

hamishforbes:
@Tehsmash linting should be fixed now 👍

FrimIdan (approved):

Tehsmash (approved):

PR #478, opened 2023-08-13
Made kubeclarity service port and targetport to dynamic

AshutoshAM2002:
## Description
I want to change kubeclarity service backend ports and targetport, but these are hardcoded, I am making it dynamic in template/service.yaml.
```yaml
spec:
  ports:
    - name: backend
      port: 8080
      protocol: TCP
      targetPort: 8080
```

- when we insert port 8080 in DNS means that our service is listening on port 8080 but port 80 is used when you don't insert any port in DNS. 
- I want the service port 80 so that we just need to browse Load Balancer DNS in the browser without giving port with DNS(DNS:8080) to access the kubeclarity UI, but service port and target port was hardcoded.
- I changed the service port from hard coded to dynamic so that we can give any port for kubeclarity service as per our convenience.

## Type of Change

[ ] Bug Fix
[ ] Breaking Change
[ ] Refactor
[ ] Documentation
[ ] Other (please describe)

## Checklist

### DIFF
--- a/charts/kubeclarity/templates/service.yaml
+++ b/charts/kubeclarity/templates/service.yaml
@@ -12,9 +12,9 @@ spec:
   type: {{ .Values.kubeclarity.service.type }}
   ports:
     - name: backend
-      port: 8080
+      port: {{ .Values.kubeclarity.service.port }}
       protocol: TCP
-      targetPort: 8080
+      targetPort: {{ .Values.kubeclarity.service.targetPort }}
     - name: runtime-scan-results
       port: {{ index .Values "kubeclarity-runtime-scan" "resultServicePort" }}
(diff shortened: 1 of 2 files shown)


FrimIdan:
@AshutoshAM2002 same comment @sambetts-cisco gave you here is still relevant https://github.com/openclarity/kubeclarity/pull/472#pullrequestreview-1560499211.
Why `targetPort` need to be configurable?
I think that in order solve your issue, only the `port` need to be configurable.

AshutoshAM2002:
Dear @FrimIdan 
Thank you for your suggestion. I agree with your point, only the port need to be configurable. I am creating another PR to make only the `port` configurable.

PR #480, opened 2023-08-14
Made kubeclarity service port to dynamic.

AshutoshAM2002:
## Description

I'd like to make the kubeclairty service's backend port configurable because we're using Load Balancer as a proxy and need to use service port 80, however your service port is hard coded with the value 8080 in template/service.yaml. I'm making it dynamic in template/service.yaml so we can use the port at our convenience.

Current service.yaml:
```yaml
spec:
  ports:
    - name: backend
      port: 8080
      protocol: TCP
      targetPort: 8080
```

After Changes in service.yaml:
```yaml
spec:
  ports:
    - name: backend
      port: {{ .Values.kubeclarity.service.port }}
      protocol: TCP
      targetPort: 8080
```

- when we insert: 8080 in DNS means that your service is listening on port 8080 but port 80 is used when you don't insert any port in DNS. I changed the service port from hard coded to dynamic.

- I have tested these changes locally they are working fine.
- I have also added a port with value 8080 in values.yaml so your running services will work fine.

Changed values.yaml:
```yaml
  service:
    type: ClusterIP
    port: 8080
    annotations: {}
```

## Type of Change

## Checklist

- [x] All code-style checks pass


AshutoshAM2002:
You're welcome, @sambetts-cisco! Glad you like the approach. BTW when can we expect merge?

PR #506, opened 2023-09-17
Update Init Container command url

alifiroozi80:
Hello
Since in the installation part, it's declared that we install the chart in the `kubeclaruty` namespace, all the services are in the `kubeclarity` namespace.
So, instead of requesting the `http://kubeclarity-kubeclarity-sbom-db:8081/healthz/ready`, we must include the namespace as well:
```shell
http://kubeclarity.kubeclarity-kubeclarity-sbom-db:8081/healthz/ready
```

---

P.S: The Kubernetes DNS structure is like this:

```shell
NAMESPACE.SVC.CLUSTER.LOCAL
```
We must use at least the `NAMESPACE.SVC` part for calling a K8s svc, so that's why we should add the `kubeclarity` namespace at the first.

### DIFF
--- a/charts/kubeclarity/templates/deployment.yaml
+++ b/charts/kubeclarity/templates/deployment.yaml
@@ -94,7 +94,7 @@ spec:
             - -c
             - >
               set -x;
-              while [ $(curl -sw '%{http_code}' "http://{{ include "kubeclarity.grype-server.name" . }}:8080/healthz/ready" -o /dev/null) -ne 200 ]; do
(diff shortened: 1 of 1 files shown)


FrimIdan:
@alifiroozi80 Thanks for your interest in the project.
I'm not sure why the PR is needed though, did you open an issue to discuss/demonstrate the problem?
AFAIK, since we are deploying all deployments in the **same** namespace there is no need to state the namespace name in the service FQDN (btw, it can be something else other than `kubeclarity`)
More info [here](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#namespaces-of-services)

If you have some issues with the kubeclarity deployment in your env, please open an issue describing the problem with as much information you can so we can help resolve it, thanks.

alifiroozi80:
You are right, thank you

Issue #498, opened 2023-09-11
Failed to analyze local images with trivy


FrimIdan:
#### What happened:

Trying to analyze local image with trivy failed
```
$ make cli && LOCAL_IMAGE_SCAN=true ANALYZER_LIST=trivy ./cli/bin/cli analyze docker.io/idanfrim/kubeclarity-cli:local
Building CLI ...
total 355312
-rwxr-xr-x@ 1 idanf  staff  181915954 Sep 11 10:25 cli
INFO[0000] Called trivy analyzer on source image docker.io/idanfrim/kubeclarity-cli:local  analyzer=trivy app=kubeclarity
ERRO[0000] failed to get image hash from sbom: repo digest property missing from Metadata.Component  analyzer=trivy app=kubeclarity
WARN[0000] "trivy" job failed: failed to get image hash from sbom: repo digest property missing from Metadata.Component  app=kubeclarity
FATA[0000] Failed to run job manager: 1 error occurred:
	* "trivy" job failed: failed to get image hash from sbom: repo digest property missing from Metadata.Component
  app=kubeclarity
```

#### What you expected to happen:

We should be able to create sbom for local images that were not pushed to registry (not repoDigest)

#### How to reproduce it (as minimally and precisely as possible):

Build a local image and DO NOT push it to remote registry.
Run `LOCAL_IMAGE_SCAN=true ANALYZER_LIST=trivy ./cli/bin/cli analyze <your local image>`

#### Anything else we need to know?:
[The fix](https://github.com/openclarity/kubeclarity/commit/89b14c11a0a072eec4df9f2ab4fa61a84ed16d71) for `syft` was to use the manifest digest instead of the repo digest, but it looks like trivy is missing the manifest digest.

#### Environment:
- KubeClarity commit 89b14c11a0a072eec4df9f2ab4fa61a84ed16d71

Tehsmash:
Just been reading this, its is a great write up of the problem with manifests and digests and why they aren't reliable: https://github.com/google/go-containerregistry/issues/895#issuecomment-753521526

It seems like the config sha / ImageID is what we want to use as this does not get effected by the existence or not existence the manifest. This is what Trivy is outputting as ImageID too IIRC. 

It would be good if local and remote copies of images could be identified as the same image.

Based on that document above repoDigests also won't be consistent between registries so today if you upload the image to two different registries so you'll get two different digests.

[...]

PR #510, opened 2023-09-27
fix: get local image hash from ImageID when RepoDigest is missing

FrimIdan:
## Description

The PR fixes an issue with local images sbom generation, when analyzing a local image the RepoDigests are missing so there is a need to use a common hash for the artifact, the code here takes the image ID (https://github.com/opencontainers/image-spec/blob/main/config.md/#imageid) as the identifier since it exists in both trivy and syft.

The PR fixes https://github.com/openclarity/kubeclarity/issues/498 and also update the changes done in https://github.com/openclarity/kubeclarity/pull/497

## Type of Change

[X] Bug Fix
[ ] New Feature
[ ] Breaking Change
[ ] Refactor
[ ] Documentation
[ ] Other (please describe)

## Checklist

- [X] I have read the [contributing guidelines](/CONTRIBUTING.md)
- [X] Existing issues have been referenced (where applicable)
- [X] I have verified this change is not present in other open pull requests
- [X] All code style checks pass
- [X] New code contribution is covered by automated tests
- [X] All new and existing tests pass

### DIFF
--- a/shared/pkg/utils/image_helper/image_helper.go
+++ b/shared/pkg/utils/image_helper/image_helper.go
@@ -41,6 +41,10 @@ type FsLayerCommand struct {
 }
 
 func GetHashFromRepoDigest(repoDigests []string, imageName string) string {
+	if len(repoDigests) == 0 {
+		return ""
+	}
+
 	normalizedName, err := reference.ParseNormalizedNamed(imageName)
 	if err != nil {
 		log.Errorf("Failed to parse image name %s to normalized named: %v", imageName, err)
@@ -225,7 +229,11 @@ func GetImageLayerCommands(imageName string, sharedConf *sharedconfig.Config) ([
 	return layerCommands, nil
 }
 
-func GetHashFromRepoDigestsOrImageID(repoDigests []string, imageID string, imageName string) string {
+func GetHashFromRepoDigestsOrImageID(repoDigests []string, imageID string, imageName string) (string, error) {
+	if imageID == "" && len(repoDigests) == 0 {
+		return "", fmt.Errorf("RepoDigest and ImageID are missing")
+	}
+
 	hash := GetHashFromRepoDigest(repoDigests, imageName)
 	if hash == "" {
 		// set hash using ImageID (https://github.com/opencontainers/image-spec/blob/main/config.md#imageid) if repo digests are missing
@@ -238,5 +246,5 @@ func GetHashFromRepoDigestsOrImageID(repoDigests []string, imageID string, image
 			hash = imageID
 		}
 	}
-	return hash
+	return hash, nil
 }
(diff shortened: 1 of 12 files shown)


shared/pkg/utils/image_helper/image_helper.go
```diff
@@ -225,14 +225,14 @@ func GetImageLayerCommands(imageName string, sharedConf *sharedconfig.Config) ([
...
-		// set hash using ManifestDigest if RepoDigest is missing
-		if idx := strings.Index(manifestDigest, ":"); idx != -1 {
-			hash = manifestDigest[idx+1:]
+		// set hash using ImageID (https://github.com/opencontainers/image-spec/blob/main/config.md#imageid) if repo digests are missing
+		if idx := strings.Index(imageID, ":"); idx != -1 {
+			hash = imageID[idx+1:]
 		} else {
-			hash = manifestDigest
+			hash = imageID
 		}
```
Tehsmash, line 236: I think we could use strings.Cut here:
```
_, remainder, found := strings.Cut(imageID, ":")
if found {
   hash = remainder
}
```

shared/pkg/utils/image_helper/image_helper.go
```diff
@@ -225,14 +225,14 @@ func GetImageLayerCommands(imageName string, sharedConf *sharedconfig.Config) ([
 	return layerCommands, nil
 }
 
-func GetHashFromRepoOrManifestDigest(repoDigests []string, manifestDigest string, imageName string) string {
+func GetHashFromRepoDigestsOrImageID(repoDigests []string, imageID string, imageName string) string {
```
Tehsmash, line 228: I wonder if this function should handle the error when len(repoDigests) == 0 and imageID == "" ? Then we could have a common error from all the scanners if we don't have the repoHash or imageID. Right now if passed imageID == "" then it'll return "" not a hash or an error.

shared/pkg/utils/image_helper/image_helper.go
pbalogh-sa: I agree, In the case of the `imageID` doesn't contain `:` we need to add
```
} else {
    hash = imageID
}
```

FrimIdan: done [`9b55c2c` (#510)](https://github.com/openclarity/kubeclarity/pull/510/commits/9b55c2cfbd846f4886b55b16e8fd8d0c9e9d4262)

shared/pkg/utils/image_helper/image_helper.go
FrimIdan: sure, make sense since we are counting on the existing of the hash - I will update the function.

pbalogh-sa:
after fixing small comments, LGTM

Tehsmash (approved):

shared/pkg/utils/image_helper/image_helper.go
FrimIdan: I've updated the function [here](https://github.com/openclarity/kubeclarity/pull/510/commits/919aeb1f4d9932d83080c8f5129d0864e1650fca), BUT I kept the old behavior of the scanners for now to basically ignore a missing hash - I've decided just to print a warning and if that will appear we will need to handle it (didn't wanted to break the current the behavior too much)
I think that it's more important when generating the sbom (and this is why we check it before) - but again, maybe we will need to also return error in the scanners flow in the future.
WDYT?

Tehsmash: Yeah happy with that, if we run into more breakages then we can address it then

PR #544, opened 2023-10-26
fix(helm): Capitalize `capabilities.drop[]` due to PSS

bauerjs1:
## Description

The [pod security standard](https://kubernetes.io/docs/concepts/security/pod-security-standards/) (PSS) level `restricted` enforces to have
```yaml
securityContext:
  capabilities:
    drop: [ALL]
```
set for all containers. At least with Kyverno, this check is case sensitive, so our Kubeclarity pods are rejected. Until now, this is set to lower case `all` in the Helm chart templates. This change capitalizes every `all` to `ALL` to be pedantically conformant with PSS.

I guess it would be more flexible to have these settings as defaults in the chart values, instead of hard-coding. If you wish, I am happy to change that, too.

## Type of Change

- [x] Bug Fix

### DIFF
--- a/charts/kubeclarity/templates/deployment.yaml
+++ b/charts/kubeclarity/templates/deployment.yaml
@@ -55,7 +55,7 @@ spec:
           securityContext:
             capabilities:
               drop:
-                - all
+                - ALL
             runAsNonRoot: true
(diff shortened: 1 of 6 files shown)


FrimIdan (approved):
LGTM

PR #727, opened 2024-08-08
docs(ci): extend PR template with before/after table

ramizpolic:
Used to test pipeline, should be removed and reopened by the PR owner


ramizpolic:
CI/CD pipeline seems to be in order

ramizpolic:
The cleanup pipeline is breaking due to PAT that we could not migrate to the new repo

PR #777, opened 2024-08-23
feat(containerruntimediscovery): add CRI-O support

zsoltkacsandi:
## Type of Change

- [ ] Bug Fix
- [X] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have read the [contributing guidelines](https://github.com/openclarity/openclarity/blob/main/CONTRIBUTING.md)
- [ ] I have verified this change is not present in other open pull requests
- [ ] All code style checks pass


containerruntimediscovery/server/crio/README.md
```diff
@@ -0,0 +1,11 @@
+In order to use CRI-O runtime with OpenClarity, you need to:
+- install `CRIU` on Kubernetes nodes
+- enable the `ContainerCheckpoint` feature gate on you cluster
+- in OpenClarity's Helm chart enable containerSecurityContext for `crDiscoveryServer`, with the following configuration:
+  ```
+  containerSecurityContext:
+    enabled: true
+    privileged: true
+    readOnlyRootFilesystem: false
```
ramizpolic, line 9: we can also add this as a comment in Helm chart itself

containerruntimediscovery/server/server.go
```diff
@@ -107,13 +107,14 @@ func (crds *ContainerRuntimeDiscoveryServer) ExportImage(c echo.Context) error {
...
-	reader, err := crds.discoverer.ExportImage(ctx, id)
+	reader, cleanup, err := crds.discoverer.ExportImage(ctx, id)
 	if err != nil {
 		if errors.Is(err, types.ErrNotFound) {
 			return echo.NewHTTPError(http.StatusNotFound, fmt.Sprintf("Image not found with ID %v", id))
 		}
 		return fmt.Errorf("failed to export image %s: %w", id, err)
 	}
+	defer cleanup()
 	defer reader.Close()
```
ramizpolic, line 118: this might cause some weird internal side effects due to call order, we should probably make it explicit:
```suggestion
	defer func() {
		_ = reader.Close() // close first as cleanup might depend on reader
		_ = cleanup()
	}
```

installation/kubernetes/helm/openclarity/templates/cr-discovery-server/daemonset.yaml
```diff
@@ -40,12 +40,32 @@ spec:
...
               readOnly: true
+            - mountPath: /var/run/crio
+              name: crio
+              readOnly: true
+            - mountPath: /var/lib/containers
+              name: varlibcontainers
+            - mountPath: /var/run/containers
+              name: varruncontainers
+            - mountPath: /tmp
+              name: tmpfs
```
ramizpolic, line 51: do we need to mount tmp here? since tmp store is only used by discoverer to save the image into, while the transfer of the image across service is actually performed via Echo server which means that we don't need tmp mount since the discoverer can write to its own tmp

containerruntimediscovery/server/server.go
zsoltkacsandi: I copied this part from the code from the ExportContainer function: https://github.com/openclarity/openclarity/blob/6e73397737cbb67423d1492d5c7866a766b092d7/containerruntimediscovery/server/server.go#L170

There is nothing wrong with the call order since deferred functions are executed in last-in-first-out order.

ramizpolic: The execution is okay, but it's clearer (for the developer) what the actual cleanup logic is when explicit

akijakya (approved):
🚀

ramizpolic (approved):

PR #807, opened 2024-08-30
fix: findings not processed

paralta:
## Description

* Increase `AssetScanProcessor ` reconcile timeout to allow time for findings processing (e.g. in AWS an asset scan with 7k findings takes around 50min to process findings)
* Fix status transition failure (from `Done` to `Pending`) that was forcing us to reconcile findings twice due to an outdated `assetScan`  

## Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/openclarity/openclarity/blob/main/CONTRIBUTING.md)

### DIFF
--- a/orchestrator/processor/assetscan/config.go
+++ b/orchestrator/processor/assetscan/config.go
@@ -23,7 +23,7 @@ import (
 
 const (
 	DefaultPollInterval     = 2 * time.Minute
-	DefaultReconcileTimeout = 5 * time.Minute
+	DefaultReconcileTimeout = 12 * time.Hour
 )
 
 type Config struct {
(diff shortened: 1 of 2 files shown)


lgecse (approved):

ramizpolic (approved):
LGTM!

Note: Although this will resolve the existing issues we have, I don't think it will prevent this behaviour from happening again. IMO I think the orchestrator component is not ready in the current state (unrelated of these changes) to handle high volume of findings. This is mainly because we dont have a way (on orchestrator, API, or DB) level to enable concurrent processing so that the next run (once the poller kicks in) processes only the findings that have not been locked by previous, still in-progress, orchestrator runs.

PR #817, opened 2024-09-04
feat(testenv): cloud e2e test improvements

akijakya:
## Description

Some minor improvements to prepare for adding cloud e2e tests to the CI pipeline (currently not planned).

## Type of Change

- [ ] Bug Fix
- [X] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [X] I have read the [contributing guidelines](https://github.com/openclarity/openclarity/blob/main/CONTRIBUTING.md)
- [X] I have verified this change is not present in other open pull requests
- [ ] All code style checks pass

### DIFF
--- a/testenv/gcp/utils.go
+++ b/testenv/gcp/utils.go
@@ -62,7 +62,7 @@ func (e *GCPEnv) afterSetUp(ctx context.Context) error {
 		User:          DefaultRemoteUser,
 		Host:          *e.serverIP,
 		Port:          utils.DefaultSSHPort,
-		LocalPort:     8080, //nolint:mnd
+		LocalPort:     8083, //nolint:mnd
 		RemoteAddress: "localhost",
 		RemotePort:    80, //nolint:mnd
 	}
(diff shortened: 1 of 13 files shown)


testenv/gcp/utils.go
```diff
@@ -62,7 +62,7 @@ func (e *GCPEnv) afterSetUp(ctx context.Context) error {
 		User:          DefaultRemoteUser,
 		Host:          *e.serverIP,
 		Port:          utils.DefaultSSHPort,
-		LocalPort:     8080, //nolint:mnd
+		LocalPort:     8083, //nolint:mnd
```
ramizpolic, line 65: not sure why we change the ports here since this is remote port

ramizpolic (approved):

testenv/gcp/utils.go
akijakya: It is even in its name that this is the local port 😃 it is used for port forwarding and if two cloud e2e test is started in parallel on the same machine, the slower would complain that the port is in use and would fail if we used the same port for all of them.

PR #982, opened 2025-01-16
fix(charts): Mount k3s containerd socket

abelfodil:
## Description

Mount k3s containerd socket such that k3s works with openclarity. 

## Type of Change

- [x] Bug Fix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/openclarity/openclarity/blob/main/CONTRIBUTING.md)

### DIFF
--- a/installation/kubernetes/helm/openclarity/templates/cr-discovery-server/daemonset.yaml
+++ b/installation/kubernetes/helm/openclarity/templates/cr-discovery-server/daemonset.yaml
@@ -40,6 +40,9 @@ spec:
             - mountPath: /var/run/containerd
               name: containerd
               readOnly: true
+            - mountPath: /run/k3s/containerd
+              name: k3s-containerd
+              readOnly: true
             - mountPath: /var/lib/docker
               name: docker
               readOnly: true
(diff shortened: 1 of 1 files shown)


installation/kubernetes/helm/openclarity/templates/cr-discovery-server/daemonset.yaml
```diff
@@ -40,6 +40,9 @@ spec:
             - mountPath: /var/run/containerd
               name: containerd
               readOnly: true
+            - mountPath: /run/k3s/containerd
+              name: k3s-containerd
+              readOnly: true
```
nymo, line 45: Sorry that I'm jumping right into this without ever contributed to this repository but I'm following this issue/problem aswell so wanted to contribute with my opinion on that topic:

Instead of mounting multiple paths it would be better if the mount path is configurable with a default value.
This will make sure that for future changes no release required. Also it will be more flexible between the systems where openclarity is being installed.

What do you think?

abelfodil: It is indeed the best approach. It might introduce breaking changes however, unless we keep this list in Values.yaml.

